PR #11, opened 2026-02-18
perf(main): debounce synchronous file writes to reduce main thread blocking

monotykamary:
## Summary

Some quick wins to avoid that "freezing" feeling when opening the clipboard history on top of the other things. Adds write debouncing to reduce blocking I/O operations on the main thread during frequent clipboard and snippet operations.

### Debounced Writes (100ms)
- **clipboard-manager.ts**: `saveHistory()` - batches clipboard history writes
- **snippet-store.ts**: `saveToDisk()` - batches snippet modification writes
- **memory.ts**: `writeLocalMemories()` - batches memory writes
- **settings-store.ts**: 
  - Added `saveSettingsAsync()` for non-blocking settings writes
  - Debounced `saveOAuthTokens()` for OAuth token persistence

### Directory Creation Optimization
- Moved `fs.existsSync` + `mkdirSync` calls from hot paths to startup:
  - `clipboard-manager`: `ensureClipboardDirs()` called once on monitor start
  - `snippet-store`: `ensureSnippetsDir()` called once on init

## Impact

- Reduces main thread blocking during rapid copy operations
- Batches multiple quick edits into single disk writes
- Improves responsiveness during heavy clipboard/snippet usage

## Testing

- TypeScript compilation passes
- All existing functionality preserved
- Debouncing only affects write timing, not behavior

### DIFF
--- a/src/main/clipboard-manager.ts
+++ b/src/main/clipboard-manager.ts
@@ -42,23 +42,27 @@ let lastClipboardText = '';
 let lastClipboardImage: Buffer | null = null;
 let pollInterval: NodeJS.Timeout | null = null;
 let isEnabled = true;
+let pendingSave: NodeJS.Timeout | null = null;
 
 // ─── Paths ──────────────────────────────────────────────────────────
 
 function getClipboardDir(): string {
-  const dir = path.join(app.getPath('userData'), 'clipboard-history');
-  if (!fs.existsSync(dir)) {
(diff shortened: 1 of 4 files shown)


monotykamary (commented):
Although logically sound (and not much to break here), there does need to be a performance harness for us to validate it instead of vibe testing it.

PR #47, opened 2026-02-19
fix(snippets): import Raycast-exported snippets and show result feedback

### LINKED ISSUE #37: [Bug] Clicking on "Import Snippet" then selecting the json exported from raycast does nothing
Steps to reproduce:
- Create a few snippets in raycast.
- Use the "Export snippets" command in raycast to export them to a json file.
- Open SuperCmd and use the import snippets command

Current: 
- Nothing happens

Expected:
- Should import all the snippets in the json

elicep01:
Fixes #37

## Root cause

Raycast exports snippets with a `"text"` field for the snippet body, but the import parser only accepted `"content"`. Every item hit the `!item.content` guard and was silently skipped — no error, no feedback.

## Changes

**`src/main/snippet-store.ts`**
- Normalise all raw items before validation: `item.content ?? item.text ?? ''` — handles Raycast exports transparently
- Deduplicate by keyword (in addition to name) to match Raycast's behaviour

**`src/renderer/src/SnippetManager.tsx`**
- Capture `{ imported, skipped }` return value from both import trigger paths (Actions menu + ⇧⌘I shortcut)
- Show a 4-second banner after import:
  - 🟢 Green: `✓ Imported 2 snippets · 1 duplicate skipped`
  - 🟡 Amber: `All 2 snippets already exist — nothing to import`

## Test plan

- [ ] Export snippets from Raycast → import via SuperCmd → green banner, snippets appear in list
- [ ] Import same file again → amber "already exist" banner, list unchanged
- [ ] Import SuperCmd's own export format → still works (no regression)
- [ ] Trigger keywords in any text field → snippets expand correctly

### DIFF
--- a/src/main/snippet-store.ts
+++ b/src/main/snippet-store.ts
@@ -386,18 +386,31 @@ export async function importSnippetsFromFile(parentWindow?: BrowserWindow): Prom
   try {
     const data = fs.readFileSync(result.filePaths[0], 'utf-8');
     const parsed = JSON.parse(data);
-
+    // Normalised shape after field remapping
     let snippetsToImport: Array<{ name: string; content: string; keyword?: string; pinned?: boolean }> = [];
 
-    // Support our export format
+    // Collect the raw items from whichever format was used
+    let rawItems: any[] = [];
     if (parsed.type === 'snippets' && Array.isArray(parsed.snippets)) {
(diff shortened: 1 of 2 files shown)


shobhit99:
tested and working!

PR #54, opened 2026-02-19
feat: besides arrow up/down, also support ctrl+j/k to cycle between items

### LINKED ISSUE #53: [FR] Allow using Ctrl-J/K hotkey besides Arrow Up/Down to cycle between items
This matches with Raycast's behavior of using `Ctrl-J/K` to cycle between items.

tuanddd:
## Summary

Add Vim-style launcher navigation: Ctrl+J / Ctrl+K now mirror Arrow Down/Up without wrap-around.

## Details

- [App.tsx:](https://github.com/SuperCmdLabs/SuperCmd/blob/main/src/renderer/src/App.tsx) refactor list navigation into a helper and wire Ctrl+J/K to the same logic as Arrow keys (no looping).
- No UI indicator is added since this is considered an advanced hotkey (also matches with Raycast's interface)

Closes #53

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -860,6 +860,22 @@ const App: React.FC = () => {
     [selectedCommand, movePinnedCommand]
   );
 
+  const moveSelection = useCallback(
+    (direction: 'up' | 'down', options: { wrap?: boolean } = {}) => {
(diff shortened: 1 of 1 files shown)


shobhit99:
Tested this, working!

PR #69, opened 2026-02-20
Add Check for Updates command

pablopunk:
- Added system command 'Check for Updates' accessible from the command bar
- Orchestrates full update flow: check → download → restart to install
- Shows toast notifications for user feedback (success/failure)

I gotta say **I have not tested** this for obvious reasons, so please double check this will work in production when there's an actual update.

CleanShot 2026-02-20 at 17 10 45

Edit: proper logo

CleanShot 2026-02-20 at 21 12 34

### DIFF
--- a/src/renderer/src/utils/command-helpers.tsx
+++ b/src/renderer/src/utils/command-helpers.tsx
@@ -13,7 +13,7 @@
  */
 
 import React from 'react';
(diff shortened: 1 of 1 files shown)


pablopunk:
thanks @shobhit99, updated!

PR #128, opened 2026-03-11
fix: keep macOS launcher on the current Space

edgyarmati:
## What changed

This changes the macOS launcher window model so normal hotkey opens use a panel-backed launcher window instead of the previous activation-heavy BrowserWindow behavior.

Key changes:
- Create the launcher as a macOS `panel` window
- Avoid forcing the launcher onto all Spaces during normal use
- Stop explicitly activating/focusing the app for normal macOS launcher opens
- Keep onboarding on the old activation path so permission/setup flows still come to the front
- Add a short README note for the macOS launcher behavior

## Why

The previous `BrowserWindow` show/focus changes were not enough to fix the Spaces bug. The launcher could still switch to the wrong desktop when opened by hotkey after Space changes.

This PR takes the narrower native-window-model approach available in Electron on macOS, which matches the successful fix path better than the abandoned `showWindow()` focus patch series.

## Compatibility impact

- macOS launcher behavior changes for normal hotkey opens: it now stays on the active Space instead of behaving like an all-Spaces overlay.
- Onboarding behavior is preserved so permission dialogs and setup still work.

## How I tested

- `npm run build`
- `npm run dev`
- Manual macOS repro from the bug report:

  1. Have Desktop 1 and Desktop 2 open, currently on Desktop 2
  2. Launch SuperCmd
  3. Open Desktop 3, then close Desktop 3
  4. Go to Desktop 1 and open SuperCmd via hotkey
  5. Confirm it stays on Desktop 1 instead of switching to Desktop 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ SuperCmd is an Electron + React launcher that focuses on Raycast extension compa
 - **Memory-aware AI** support (Supermemory integration)
 - AI integrations (Ollama / OpenAI / Anthropic / ElevenLabs for speech)
 - Native macOS helpers (hotkeys, color picker, speech, snippet expansion)
+- macOS launcher panel behavior that stays on the active Space for normal hotkey opens
 
 ## Tech Stack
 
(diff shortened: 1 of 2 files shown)


shobhit99:
@EdGy2k This works perfectly! tested it out. Thanks!

PR #195, opened 2026-04-10
feat: clipboard history quick-paste (Cmd+1-9) and fast native paste

ericmason:
## Summary

- **Cmd+1 through Cmd+9** instantly pastes the Nth item from clipboard history (Alfred-style workflow)
- **⌘ number badges** shown on the first 9 items so users know which key to press
- **Escape dismisses the window** entirely instead of returning to the main launcher
- **~8x faster paste** via a native N-API addon that replaces osascript process spawning with in-process CGEvent posting (~50ms vs ~400ms)

## Motivation

Users migrating from Alfred expect number-key quick-paste in clipboard history. The existing arrow-key navigation is slower for muscle-memory workflows. The paste itself was also slow due to spawning osascript twice with a 200ms sleep — this is now handled by a native addon running inside the Electron process.

### UI changes (commit 1)
- `ClipboardManager.tsx`: Cmd+1–9 keydown handler + `⌘N` kbd badges on items 1–9
- `App.tsx`: Escape calls `window.electron.hideWindow()` instead of returning to launcher

### Native addon (commit 2)
- `src/native/fast-paste-addon/fast_paste.mm`: N-API addon with `activateAndPaste(bundleId)` — activates via `NSRunningApplication`, polls until frontmost, posts ⌘V via `CGEvent`
- Runs inside Electron process, inherits accessibility permissions
- Falls back to original osascript path if addon fails to load
- Auto-detects arch (arm64/x64) in build script

## Test plan
- [x] Cmd+1 through Cmd+9 pastes correct item into target app
- [x] Number badges display correctly on first 9 items
- [x] Escape dismisses window (does not show main launcher)
- [x] Paste works with text, URLs, and images
- [x] Fallback to osascript works if addon is missing
- [x] `npm run build` completes without errors


shobhit99:
Hi @ericmason tested this out. the pasting is instant now.

two things
1. Can we put the cmd + n options on the right side of the clipboard items, currently it's on the left side. this would make the design consistent
2. Can you reuse the same logic for snippets and AI Inline prompt paste.

shobhit99:
@claude can you create a new branch from `ericmason:feat/clipboard-history-improvements` work on the suggested changes by me above

shobhit99:
I am merging this, will add the requested changes in new commit

PR #196, opened 2026-04-10
Whisper performance + fast-paste migration + developer UX

ericmason:
## Summary

This branch lands a set of whisper + paste-performance improvements plus the dev-UX changes needed to test them reliably.

Core feature work:

- **Persistent whisper.cpp server + native fast-paste for dictation** (initial commits: Metal GPU backend, native mic capture with `getUserMedia` fallback, FFT waveform visualization, initial prompt setting, double-tap resend, model size selector).
- **Finish the fast-paste migration** — every paste/type path (clipboard history, snippet paste, snippet keyword auto-expand, `prompt-apply-generated-text`, `replace-live-text`, `paste-file`, `paste-text`) now uses the native CGEvent addon instead of `osascript`. osascript preserved as a safety-net fallback.

Bundled bug fixes and dev UX:

- **Fix — app activation in native addon**: `activateWithOptions:0` is a no-op on modern macOS when the caller isn't already frontmost, which we aren't after `mainWindow.hide()`. Now uses `NSApplicationActivateIgnoringOtherApps`, so ⌘V actually fires into the previously-frontmost app instead of whatever macOS focused next.
- **Fix — DevTools shortcut scope**: `⌘⌥I` was registered via `globalShortcut.register` and intercepted the combo system-wide in dev mode. Scoped to SuperCmd's own \`webContents\` via `before-input-event` — Chrome/VS Code `⌘⌥I` works normally again.
- **Dev UX — `open -n` launch scripts**: `./node_modules/.bin/electron .` attributes TCC to whatever terminal/SSH daemon spawned it. `npm run dev:gui` uses `open -n --env ... -a Electron.app` so launchd roots TCC responsibility in Electron itself and the Accessibility prompt correctly says \"Electron.\"
- **Feat — onboarding re-check**: if the user has seen onboarding but is missing a core permission (Accessibility or Input Monitoring), the launcher reopens at the Permissions step instead of replaying all 7 intro steps. Handles the common case of reinstalling Electron in dev and losing the per-binary TCC grant.
- **Whisper stability**: extra logging around hotkey registration, hold watcher, and type-live; `whisper-speak-toggle` debounce; first-launch auto-open is now gated on `!hasSeenOnboarding`.

## Whisper indicator: never steal focus (latest update)

Added after initial review to fix a long-standing side effect where showing the whisper overlay blurred the previously-frontmost app:

- **Detached whisper / memory-status overlays are now fully passive on macOS**: created with `type: 'panel'` + `show: false`, then surfaced via `childWindow.showInactive()` on `ready-to-show`. The default `show: true` path routes through `makeKeyAndOrderFront:` and triggers `[NSApp activate]` even with `focusable: false`, which was blurring launcher-style apps the user had open and causing them to hide-on-blur.
- **Removed the 50/180/360 ms `activateLastFrontmostApp()` focus-restore loop** that ran whenever a detached overlay opened with `preserveFocusWhenHidden`. With the inactive panel the overlay never takes focus, so the \"restore\" loop was itself re-ordering windows in other apps for no reason.
- **Shortened the whisper double-tap resend window from 3000 ms → 300 ms** so a real double-tap reads as resend and a casual second trigger doesn't replay the last transcript.

## Why the paste migration matters

The `osascript` paste path has three problems:

1. **Slow** — ~200–300ms per `osascript` spawn.
2. **Permission cliff** — requires `Automation → System Events`, a separate TCC grant that many users reject. The prompt text is unhelpful.
3. **Silent failure** — error 1002 (\"osascript is not allowed to send keystrokes\") is swallowed and produces no user-visible feedback.

The native CGEvent path only needs Accessibility (which users expect for a launcher) and is roughly an order of magnitude faster. Every leaf function now tries the fast path first and falls back to osascript only if the addon is unavailable or fails — so behavior is strictly a superset of what was there before.

## Test plan

- [ ] First-launch onboarding still appears for fresh installs
- [ ] Revoke Accessibility in System Settings, relaunch → onboarding opens at Permissions step
- [ ] Clipboard history → Enter pastes into previously-frontmost app
- [ ] Snippet paste via Enter works
- [ ] Snippet keyword auto-expand (e.g. \`;email\`) works
- [ ] Whisper dictation types into target app (both streaming and final paste)
- [ ] Triggering the whisper overlay while another launcher-style app is frontmost does **not** cause that app to hide
- [ ] Double-tapping the whisper hotkey within 300 ms resends the last transcript; a slower second tap starts a fresh recording
- [ ] Global AI prompt \"apply generated text\" replaces selection
- [ ] `⌘⌥I` in SuperCmd launcher opens SuperCmd DevTools
- [ ] `⌘⌥I` in Chrome/VS Code opens their DevTools (not SuperCmd)


shobhit99:
integrated the fast paste from this. closing this PR

PR #197, opened 2026-04-10
Boost recently used commands in search results

ericmason:
## Summary
- Adds an Alfred-style recency boost to search scoring: recently launched commands rank higher when search scores are close
- Uses the existing `recentCommands` tracking (already persisted to settings) — no new state or storage needed
- Most recently used command gets +80 points, decaying quickly (`80 / (1 + rank)`), so it nudges results without overriding genuinely better matches
- Also uses recency as a secondary tiebreaker when scores are exactly equal

**Example:** Searching "chrome" after launching Google Chrome now puts it above Chrome Remote Desktop, even though the latter has a prefix match advantage.

## Test plan
- [ ] Launch a command (e.g. Google Chrome), then search for a term that matches multiple commands — the just-launched one should appear higher
- [ ] Verify that exact title/alias matches still rank above recency-boosted results
- [ ] Verify the "Recent" section on the home screen is unaffected

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -1359,8 +1359,8 @@ const App: React.FC = () => {
   const calcOffset = calcResult ? 1 : 0;
   const contextualCommands = commands;
   const filteredCommands = useMemo(
-    () => filterCommands(contextualCommands, searchQuery, commandAliases),
-    [contextualCommands, searchQuery, commandAliases]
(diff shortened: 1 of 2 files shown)


shobhit99:
Hi @ericmason this is great, 
can you also make it persist, so it's not reset on restarts

shobhit99:
This was added in 1.0.25, closing the PR

PR #200, opened 2026-04-10
fixed the geneis search book extension

shobhit99:


shobhit99:
@claude please review this PR and see if this will cause any regression

PR #202, opened 2026-04-10
Added auto-close setting for whisper bar and fixed warmup banner

ericmason:
## Summary
- Adds an **Auto-Close After Dictation** toggle in Settings > AI > Whisper (on by default)
  - When enabled, the whisper bar automatically closes after dictation finishes instead of lingering idle on screen
  - When disabled, preserves the existing behavior (bar stays open for back-to-back dictation)
- Fixes the **warmup banner flashing** on every dictation start — the "Loading models..." banner now only appears if warmup actually takes >200ms, so it's invisible when the server is already warm

## Changes
- `electron.d.ts` — added `whisperAutoClose` to `AISettings`
- `en.json` — added i18n strings for the new toggle
- `AITab.tsx` — added the toggle UI in the Whisper settings panel
- `App.tsx` — loads and passes the `whisperAutoClose` setting to `SuperCmdWhisper`
- `SuperCmdWhisper.tsx` — respects `autoClose` prop in push-to-talk and toggle flows; delays warmup banner with a 200ms timer

## Test plan
- [ ] Enable auto-close (default) → dictate → bar should close after text is typed
- [ ] Disable auto-close → dictate → bar should stay open in idle state
- [ ] With server already warm: no "Loading models..." flash on dictation start
- [ ] Cold start (first dictation after app launch): "Loading models..." banner should appear after ~200ms if warmup is still in progress
- [ ] Toggle persists across app restarts

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -367,6 +367,7 @@ const App: React.FC = () => {
   const [onboardingRequiresShortcutFix, setOnboardingRequiresShortcutFix] = useState(false);
   const [onboardingHotkeyPresses, setOnboardingHotkeyPresses] = useState(0);
   const [launcherShortcut, setLauncherShortcut] = useState('Alt+Space');
+  const [whisperAutoClose, setWhisperAutoClose] = useState(true);
   const [showActions, setShowActions] = useState(false);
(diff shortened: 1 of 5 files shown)


shobhit99:
tested this, works!

PR #203, opened 2026-04-10
Fix inline AI prompt deleting content before selection and duplicate streaming

allgemeiner-intellekt:
- Fix 1: Inline AI prompt deleting content before selection — When applying a inline AI prompt result, the code sent prev.length backspaces to clear the selected text. But since the selection was still active, the first backspace already deleted the entire selection, and every subsequent one ate a character before it. Fixed by sending a single backspace in both replaceTextDirectly and replaceTextViaBackspaceAndPaste.

- Fix 2: AI streaming output duplicating every word (which is found when testing fix 1) — AI stream event listeners (onAIStreamChunk/Done/Error) were registered in useEffect hooks without cleanup functions, causing listeners to stack on re-renders. Each chunk was processed multiple times, doubling every word in the output. Fixed by returning removal functions from the preload bridge and calling them in useEffect teardown across useCursorPrompt, useAiChat, and PromptApp.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -7761,11 +7761,13 @@ async function replaceTextDirectly(previousText: string, nextText: string): Prom
 
   try {
     if (prev.length > 0) {
+      // The original selection is still active in the target app, so a single
+      // backspace (or simply typing) replaces the entire selection.  Sending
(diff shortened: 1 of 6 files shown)


shobhit99:
Hi @allgemeiner-intellekt Thanks! tested this out. works fine

PR #204, opened 2026-04-10
Added whisper.cpp model size selection

ericmason:
## Summary
- Adds a **Model Size** dropdown in Settings > AI > Whisper for whisper.cpp: tiny (~39MB), base (~140MB), small (~466MB), medium (~1.5GB), large (~2.9GB)
- Red validation border + warning message when the selected model hasn't been downloaded yet
- Model size is cached in memory to avoid repeated disk reads on every transcription call
- Cache is invalidated when the setting changes

## Changes
- `settings-store.ts` — added `whisperCppModelSize` to `AISettings` interface and defaults
- `main.ts` — replaced hardcoded `WHISPERCPP_MODEL_NAME` with cached `getWhisperCppModelSize()`, captured size once in download function to prevent mid-download inconsistency
- `electron.d.ts` — added `whisperCppModelSize` to renderer-side type
- `AITab.tsx` — model size dropdown with red border validation for undownloaded models
- `en.json` — i18n strings with model descriptions and file sizes

## Test plan
- [ ] Select each model size — verify the download card reflects the correct model
- [ ] Select a model that isn't downloaded — red border and warning appear
- [ ] Download a non-default model — verify it downloads the correct size
- [ ] Dictate with different model sizes — verify transcription works
- [ ] Restart app — verify the model size selection persists

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -161,8 +161,27 @@ function getNativeBinaryPath(name: string): string {
 }
 
 const WHISPERCPP_FRAMEWORK_VERSION = 'v1.8.3';
-const WHISPERCPP_MODEL_NAME = 'base';
-const WHISPERCPP_MODEL_URL = `https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-${WHISPERCPP_MODEL_NAME}.bin`;
+const WHISPERCPP_DEFAULT_MODEL_SIZE = 'base';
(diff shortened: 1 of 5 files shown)


shobhit99:
Hi @ericmason tested this out. the model selection works. there's one issue tho. downloading large model returns 404

Also can you add an option to delete the model as well. people might try out the 4gb version if it's too slow they might want to free up the space

image

shobhit99:
@ericmason can you fix the conflict and the 404 issue, also another user had made a change for checking the sha hash for the downloaded model with the static hash we have in code. can you also add that for all of these models

PR #233, opened 2026-04-13
Improve launcher search rendering and pinyin matching

murongg:
## Summary
- extract launcher command filtering into an indexed search helper to reduce per-keystroke work
- keep visible launcher results synchronized with the current query and clear stale async calculator state
- stabilize the launcher result list paint/layout in the glass surface to reduce transient stale-row artifacts
- add pinyin and initials matching for Chinese launcher commands such as app names

## Test Plan
- npm run build:renderer
- spot-check `weixin` / `wxkf` matches `微信开发者工具`


murongg:
Why did Claude's review fail?

shobhit99:
@murongg is there any alternative to pinyin. the unpacked size is almost 1mb

murongg:
> @murongg is there any alternative to pinyin. the unpacked size is almost 1mb

This kind of library seems to be quite large.

shobhit99:
@murongg does pinyin support other languages, we have embeded the new calculator which supports calculation in multiple languages. it would be great if can do all in one library

murongg:
> @murongg does pinyin support other languages, we have embeded the new calculator which supports calculation in multiple languages. it would be great if can do all in one library

Pinyin is a Latin alphabet phonetic system used to represent the pronunciation of modern standard Mandarin Chinese. It is only valid for Chinese.

shobhit99:
@murongg can you make use of "transliteration" here, it will work with other languages as well

shobhit99:
@murongg I have used the transliteration, this is the branch name `transliteration-integration` can you see if it works for you and solves your use case?

murongg:
> @murongg I have used the transliteration, this is the branch name `transliteration-integration` can you see if it works for you and solves your use case?

ok, but i not found `transliteration-integration`

shobhit99:
@murongg the transliteration integration is live in main branch, and in version 1.0.17

PR #362, opened 2026-05-06
feat: add suport for importing data from raycast exports

### LINKED ISSUE #284: Import raycast extensions
Would be nice to have!

brrock:
fixes: #284 
## Summary

This PR adds first-class import support for Raycast backup files (`.rayconfig`) and fills in the storage gaps that were blocking a believable migration into SuperCmd.

The importer now runs in the Electron main process. It can open a Raycast backup, prompt for the backup password, decrypt it locally, preview what it found, and import the parts of the backup that SuperCmd can represent safely today.

## What imports now

- Raycast settings that have direct SuperCmd equivalents
- global launcher hotkey
- command hotkeys where a Raycast command, app, or script can be mapped onto a SuperCmd command
- disabled extension commands
- disabled script commands
- quicklinks
- snippets
- notes
- installed Raycast Store extensions
- extension preferences
- script command folder paths
- pinned or favorite commands when Raycast exports them in a usable form
- AI chat history

## AI chat handling

This PR now imports Raycast AI chats instead of skipping them.

The important part is that this does not require a large UX rewrite. SuperCmd already had a usable AI chat interface, but its history lived entirely in renderer `localStorage`. That made import awkward and brittle. This PR introduces a shared main-process AI chat store, migrates old local chat history into it on startup, and updates the existing AI chat UI to read and write through that shared store.

That means imported Raycast conversations simply appear inside the current SuperCmd AI chat history like normal conversations.

The importer maps Raycast chat data into SuperCmd's existing conversation model:

- conversation id
- title
- message list
- created and updated timestamps
- optional source metadata

To avoid unnecessary UI churn, Raycast-specific structures are flattened into the current `user` / `assistant` transcript model. So the chat history itself is preserved, but Raycast-only affordances such as richer tool-call rendering are not recreated in the UI.

## Storage refactor

This PR also continues the storage cleanup needed to make import reliable:

- extension preferences now live in a main-process store instead of renderer-only `localStorage`
- AI chats now live in a main-process store instead of renderer-only `localStorage`
- legacy extension prefs are merged forward on startup
- legacy `sc.aiChat.conversations` data is merged forward on startup

This keeps existing SuperCmd users' data intact while giving the importer a real place to write imported state.

## Script command import

Script command migration is more complete now:

- imported Raycast script-command folders are persisted in SuperCmd settings
- the script runner reads from those folders
- Raycast's disabled script-command entries are resolved from their path-based identifiers onto SuperCmd script command ids so disabled state survives import

## Aliases, favorites, and hotkeys

Raycast command hotkeys are imported when they can be mapped onto known SuperCmd commands.

Favorites are imported when Raycast exposes them through `pinnedMenuItems`. In backups where that array is empty, there is simply nothing to import.

Aliases remain best-effort. Raycast does not appear to expose a clean first-class alias model in this backup format, so alias import is derived conservatively from exported search-term data only when it looks intentional rather than noisy.

## MCPs

MCP server import is still intentionally out of scope.

Raycast can export MCP server data, but SuperCmd does not yet have a normalized persisted MCP configuration model that this importer can safely target. Importing MCPs now would mean either guessing at storage shape or pushing credential-adjacent config into an unstable format.

So the current behavior is deliberate:

- the importer detects exported MCP servers
- it reports them as skipped
- it does not silently drop them
- it does not import them into a guessed schema

To support MCP import properly, SuperCmd would first need:

1. a canonical persisted MCP server config model
2. explicit mapping from Raycast MCP fields onto that model
3. a policy for secrets and auth material
4. a decision on whether imported MCP servers should be disabled by default until reviewed

## Still out of scope

- clipboard history import
- MCP server import
- perfect one-to-one alias migration from Raycast
- full builtin-command parity for every Raycast builtin
- Raycast-specific AI metadata beyond what fits cleanly into SuperCmd's current conversation UI

## Verification

- `npm run build:main`
- `npm run build:renderer`

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ SuperCmd is an Electron + React launcher that focuses on Raycast extension compa
 ## Key Features
 
 - **Raycast extension compatibility** — `@raycast/api` and `@raycast/utils` shims; install extensions directly from the Raycast store
+- **Raycast backup import** — import encrypted `.rayconfig` backups with settings, hotkeys, extensions, scripts, quicklinks, snippets, notes, and extension prefs
 - **AI cursor prompt** — inline AI suggestions at your cursor position across any app
 - **AI chat** — chat with configurable providers (OpenAI / Anthropic / Ollama / Gemini / OpenAI-compatible)
 - **Hold-to-speak dictation** — Wispr Flow-style voice input; hold hotkey, speak, release to type (Whisper, Parakeet, or native macOS STT)
@@ -131,6 +132,31 @@ SuperCmd needs the following permissions. The app will prompt you on first use, 
 ### Auto-updates
 
 SuperCmd includes a built-in auto-updater backed by GitHub Releases. You can check for updates manually by searching "Check for Updates" in the launcher, or install a downloaded update on next launch.
+
+### Raycast Backup Import
+
+SuperCmd can import encrypted Raycast `.rayconfig` backups from the General settings tab.
+
+It currently imports:
+- Raycast settings that map cleanly to SuperCmd
+- the global launcher hotkey
+- command hotkeys
+- quicklinks
+- snippets
+- notes
+- installed Raycast extensions
+- extension preferences
+- script command folders
+- disabled script commands
+- disabled extension commands
+
+It intentionally skips or only partially maps:
+- AI chats
+- clipboard history
+- MCP server config
+- Raycast aliases, where the backup does not expose a clean first-class field
+
+The importer decrypts backups locally and prompts for the backup password before reading the file.
 
 ---
 
--- a/src/main/ai-chat-store.ts
+++ b/src/main/ai-chat-store.ts
@@ -1,0 +1,248 @@
+import { app } from 'electron';
+import { createHash } from 'crypto';
+import * as fs from 'fs';
+import * as path from 'path';
+
+export interface AiChatMessageRecord {
+  id: string;
+  role: 'user' | 'assistant';
+  content: string;
+  createdAt: number;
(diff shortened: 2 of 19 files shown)


brrock:
This is very early and hasn't been fine tuned fully. However I fully invite you to nitpick, review and test!

brrock:
@codex

shobhit99:
Thanks for working on this, i was about to start on this. allow me some time. will thoroughly test this.
ignore the claude action failure

shobhit99:
@brrock I couldn't find the option in General / Advanced tab to import the config file. can you see if there's anything missing here

brrock:
image
it should be in general

brrock:
@codex

brrock:
@codex

shobhit99:
@brrock Few changes
1. While the importing is going, we don't have any visibility on progress. it would be great if we can add the current step on UI
2. on the password screen, we should add a short message for where you can find the password
3. let's move the backup to advanced tab
4. Once the import is completed, the hotkeys are mapped correctly but not activated. the app needs a restart for it to work

brrock:
> @brrock Few changes
> 
> 1. While the importing is going, we don't have any visibility on progress. it would be great if we can add the current step on UI
> 
> 2. on the password screen, we should add a short message for where you can find the password
> 
> 3. let's move the backup to advanced tab
> 
> 4. Once the import is completed, the hotkeys are mapped correctly but not activated. the app needs a restart for it to work

Sounds good. Already implemented progress and yeah we could restart the app after import.

PR #363, opened 2026-05-06
Performance improvements: Cut latency on show + Enter; fire-and-forget all launch paths

cmer:
## Summary

Targets two perceived-latency bugs:

- **Slow show after global hotkey** — the show path was awaiting a ~50 ms window-management worker IPC before `mainWindow.show()`, the `window-shown` IPC fired before the OS painted the window, and the renderer ran heavy state-resets + a per-show `fetchCommands` (full ~1000-item reconciliation) before focusing the input. Fixed by reordering the show path, dropping the await, skipping `fetchCommands` when the cache is fresh, focusing the input synchronously, and deferring non-critical post-show work to `requestIdleCallback`. Also disabled `backgroundThrottling` on the launcher window so the first frame after `show()` isn't catch-up.
- **Enter felt slow / sometimes ignored** — every launch primitive (`openAppByPath`, `openSettingsPane`, `openTargetWithApplication`, `shell.openPath`, `shell.openExternal`, `browser-search-history.openInDefaultBrowser`, the script-commands folder IPCs) was awaited, holding the renderer's IPC await — and therefore the launcher window — for the full LaunchServices dispatch (commonly 1–3 s on macOS for first launch / Gatekeeper checks). Now all of those are fire-and-forget; the launcher hides on the existing 50 ms timer regardless of the launch's own duration. Added an in-flight guard on `handleCommandExecute` so a fast double-Enter no longer re-fires the same command.

## Test plan

- [ ] Trigger the global hotkey ten times in a row from a cold state — launcher should appear within one frame each time.
- [ ] Type a few characters and press Enter once — must execute on the first press, every time.
- [ ] Press Enter twice rapidly on a slow command — the second press must be ignored (no double-launch).
- [ ] Launch Microsoft Word (or any app with slow first-launch) — launcher must hide instantly while Word starts in the background.
- [ ] Open a System Settings pane (Display, About) — launcher must hide instantly even when System Settings is already running.
- [ ] Run a quicklink (URL, file path, app target) — launcher must hide instantly.
- [ ] Run a browser search — launcher must hide instantly.
- [ ] Window-management presets must still target the previously-frontmost window correctly (verifies the lazy WM capture in commit \`0ffc5b1\` / \`b606ee8\`).
- [ ] Install or uninstall an extension — list refreshes via the existing \`commands-updated\` IPC, confirming the per-show fetch skip didn't break staleness recovery.

### DIFF
--- a/src/main/browser-search-history.ts
+++ b/src/main/browser-search-history.ts
@@ -192,12 +192,11 @@ export async function openInDefaultBrowser(rawInput: string): Promise<{
 }> {
   const resolved = resolveInput(rawInput);
   if (!resolved) return { ok: false, resolved: null };
-  try {
-    await shell.openExternal(resolved.url);
-  } catch (e) {
+  // Fire-and-forget: don't await LaunchServices. The renderer's IPC await
+  // would otherwise hold the launcher visible for the full dispatch window.
+  void shell.openExternal(resolved.url).catch((e) => {
     console.error('Failed to open URL in default browser:', e);
-    return { ok: false, resolved };
-  }
+  });
   recordEntry(rawInput.trim(), resolved);
   return { ok: true, resolved };
 }
--- a/src/main/commands.ts
+++ b/src/main/commands.ts
@@ -1189,6 +1189,9 @@ async function openAppByPath(appPath: string): Promise<void> {
     detached: true,
     stdio: 'ignore',
   });
+  child.on('error', (err) => {
+    console.error(`Failed to open app: ${appPath}`, err);
+  });
   child.unref();
 }
 
@@ -1204,7 +1207,11 @@ async function openSettingsPane(identifier: string): Promise<void> {
   const url = identifier.startsWith('com.apple.')
     ? `x-apple.systempreferences:${identifier}`
     : `x-apple.systempreferences:com.apple.settings.${identifier}`;
-  spawn('/usr/bin/open', [url], { detached: true, stdio: 'ignore' }).unref();
+  const child = spawn('/usr/bin/open', [url], { detached: true, stdio: 'ignore' });
+  child.on('error', (err) => {
+    console.error(`Failed to open settings pane: ${url}`, err);
+  });
+  child.unref();
 }
 
 // ─── Public API ─────────────────────────────────────────────────────
--- a/src/main/commands.ts
+++ b/src/main/commands.ts
@@ -1179,12 +1179,10 @@ async function discoverSystemSettings(): Promise<CommandInfo[]> {
 // ─── Command Execution ──────────────────────────────────────────────
 
 async function openAppByPath(appPath: string): Promise<void> {
-  // open(1) on macOS dispatches the launch to LaunchServices and is supposed
-  // to return quickly, but it can block for 1-3 s on first launch (Gatekeeper
-  // signature checks, sealed-package validation, LaunchServices metadata
-  // rebuilds — Microsoft Office is a frequent offender). Awaiting it kept
-  // the launcher visible the whole time and made the launch feel slow even
-  // though the app's own startup is identical either way. Fire-and-forget.
+  // open(1) is supposed to return quickly after dispatching to
+  // LaunchServices, but can block 1-3s on first launch (Gatekeeper,
+  // sealed-package validation — Microsoft Office is a frequent offender).
+  // The launch is dispatched async either way, so fire-and-forget.
   const child = spawn('/usr/bin/open', [appPath], {
     detached: true,
     stdio: 'ignore',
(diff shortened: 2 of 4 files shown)


shobhit99:
## Code Review

The perf work here is solid — reordering `show()` before `window-shown`, focus-first in `onWindowShown`, `requestIdleCallback` deferral, fire-and-forget launch paths, and the in-flight guard on `handleCommandExecute` are all good calls. Two critical issues before this lands.

---

### [CRITICAL] Version regression — `1.0.22` → `1.0.21`

**`package.json` line 3**

`main` is currently at `1.0.22`. This PR downgrades to `1.0.21`. Auto-updaters compare semver: any user already running 1.0.22 will never receive this build as an update, and publishing 1.0.21 to releases after 1.0.22 shipped breaks the release channel for everyone who hasn't auto-updated yet.

The version should be bumped forward (1.0.23 or whatever the next intended release is), not rolled back.

---

### [CRITICAL] Existing LM Studio users get silently broken AI — no migration

**`src/main/settings-store.ts`, `src/main/ai-provider.ts`**

Users who previously configured LM Studio will have `provider: "lm-studio"` persisted in `settings.json`. The `hydrateAISettings` spread (`{ ...DEFAULT_AI_SETTINGS, ...(raw || {}) }`) preserves that value in memory as-is — `loadSettings` has no migration that maps `lm-studio` → `openai-compatible`.

After this PR ships, `streamAI` and `streamAIChat` reach their `switch (route.provider)` with `provider === 'lm-studio'`, find no matching case, and return an empty generator. No error is thrown, no warning is shown — AI silently produces nothing.

Fix: add a one-shot migration in `hydrateAISettings` or `loadSettings`:

```ts
if ((merged.provider as string) === 'lm-studio') {
  merged.provider = 'openai-compatible';
  merged.openaiCompatibleBaseUrl = merged.openaiCompatibleBaseUrl || (merged as any).lmStudioBaseUrl || 'http://127.0.0.1:1234/v1';
  merged.openaiCompatibleApiKey  = merged.openaiCompatibleApiKey  || (merged as any).lmStudioApiKey  || '';
  merged.openaiCompatibleModel   = merged.openaiCompatibleModel   || (merged as any).lmStudioModel   || '';
}
```

A toast on next launch ("LM Studio has been migrated to Custom (OpenAI-compatible)") would help affected users understand why their config changed.

---

### [Informational] `openInDefaultBrowser` always returns `ok: true` on failure

**`src/main/browser-search-history.ts` line 201**

Before: `shell.openExternal` failure returned `{ ok: false, resolved }` and skipped `recordEntry`. After: errors are fire-and-forget, `recordEntry` is always called, and the function always returns `{ ok: true, resolved }`. Callers that check the `ok` field are now misled; the history will record entries that were never actually opened. Low severity (this is a detail view), but worth noting if the caller uses `ok` to decide UX.

---

### [Informational] Wrong error attributed in URL fallback log

**`src/main/main.ts` `openTargetWithApplication`**

```js
void shell.openExternal(externalTarget).catch((error: unknown) => {
  if (externalTarget !== rawTarget) {
    void shell.openExternal(rawTarget).catch(() => {
      console.error(`Failed to open URL: ${rawTarget}`, error); // `error` is from externalTarget, not rawTarget
    });
    return;
  }
  ...
});
```

When both `externalTarget` and `rawTarget` fail, the inner catch logs `externalTarget`'s error object against `rawTarget`'s string. Not a runtime bug, just a misleading log. The inner catch should capture its own error: `.catch((rawErr) => console.error(..., rawErr))`.

---

### [Informational] `backgroundThrottling: false` — permanent battery trade-off

**`src/main/main.ts` `createWindow`**

Disabling background throttling on the launcher window means Chromium never deprioritises the renderer's task queue while the window is hidden — which is most of the app's lifetime. The gain (snappy first-frame after `show()`) is real, but so is the CPU cost on battery. Worth measuring on an M-series MacBook under a workload that triggers the global hotkey frequently.

shobhit99:
## Corrected Review (branch not rebased with main)

Ignore my previous comment — it was based on a stale `git diff origin/main` which picked up unrelated changes from main (LM Studio removal, version bump) that aren't part of this PR. True diff is 4 files, all perf work. Reanalyzed below.

---

### [INFORMATIONAL] `openInDefaultBrowser` always returns `ok: true` on failure

**`src/main/browser-search-history.ts` line 201**

The fire-and-forget refactor also changes the failure contract: errors are swallowed, `recordEntry` is always called, and the function now always returns `{ ok: true, resolved }`. A URL that failed to open gets recorded in browser search history. Low impact if the IPC caller doesn't branch on `ok`, but worth knowing.

---

### [INFORMATIONAL] Wrong error attributed in URL fallback log

**`src/main/main.ts` `openTargetWithApplication`**

```js
void shell.openExternal(externalTarget).catch((error: unknown) => {
  if (externalTarget !== rawTarget) {
    void shell.openExternal(rawTarget).catch(() => {
      console.error(`Failed to open URL: ${rawTarget}`, error); // `error` is from externalTarget
    });
    return;
  }
  ...
});
```

When `externalTarget` and `rawTarget` both fail, the inner `.catch(() => {})` swallows the rawTarget error and logs `externalTarget`'s error object against `rawTarget`'s string. Not a runtime bug, just a misleading log. Fix: `.catch((rawErr) => console.error(..., rawErr))`.

---

Everything else looks correct — the `showWindow` reordering, WM capture race fix (`resolveLauncherEntryTargetWorkArea` fallback), `requestIdleCallback` deferral with the 5-min TTL guard, the `executingCommandRef` in-flight guard (properly cleared in `finally`), and the fire-and-forget spawn paths are all solid.

PR #365, opened 2026-05-07
(feat) LM Studio provider

### LINKED ISSUE #173: Support for LMSutdio
Support for LMSutdio

NZ-Linix:
Adding support for LM Studio as a provider:
- Selectable "LM Studio" in the LLM config page (next to Ollama, Custom OpenAI compatible, etc.)
- Ability to set a server base URL (defaults to local LM Studio at `http://127.0.0.1:1234/v1`)
- Ability to set an optional API Key (by default hidden as LM Studio usually doesn't require an API Key) 
- Ability to choose a model (fetches a list from LM Studio) or when using Auto it uses the currently loaded model (if none is loaded it auto-loads the last used model)

I also added translations for: German, English, Spanish, French, Italian, Japanese, Korean, Chinese (Simplified & Traditional)

Closes #173

image

### DIFF
--- a/src/main/ai-provider.ts
+++ b/src/main/ai-provider.ts
@@ -32,7 +32,7 @@ export interface AIChatRequestOptions {
 // ─── Model routing ────────────────────────────────────────────────────
 
 interface ModelRoute {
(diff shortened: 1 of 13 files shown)


shobhit99:
Tested it out, works! Thanks!

PR #366, opened 2026-05-07
Update homebrew formula each time a new tag is created

rishabhj1717:
@shobhit99 This will require a private token with the following access to be created before being merged into main

Create a fine scoped github PAT token
Token Name: `HOMEBREW_TAP_TOKEN`
Scope
`contents: write` and `pull-requests: write` on `SuperCmdLabs/homebrew-supercmd`


rishabhj1717:
@shobhit99 can you review this PR?

PR #373, opened 2026-05-08
fix: correct settings toggle knob alignment for all font sizes

monotykamary:
### Problem

Settings toggle sliders (AITab, ExtensionsTab) had a broken "on" position where the knob overshot the pill slightly (in the `small` font size), making it look inconsistent with the "off" position.

| Before | After |
|-|-|
| image |  image |

### Root cause

The "off" position used `left-0.5` (0.125rem) but the "on" position used a hardcoded `translate-x-[18px]`. Since the app supports dynamic root font sizes (12–20px), this fixed pixel value did not scale with the pill width, causing misalignment at **different font sizes**.

### Fix

Replaced the asymmetric `left-0.5` / `translate-x-[18px]` approach with a symmetric `left-0.5` ↔ `right-0.5` toggle. This guarantees identical inner padding on both sides at any font size, keeping the knob perfectly flush inside the pill.

### Files changed

- `src/renderer/src/settings/AITab.tsx` (2 toggle instances)
- `src/renderer/src/settings/ExtensionsTab.tsx` (emoji picker toggle)

### DIFF
--- a/src/renderer/src/settings/AITab.tsx
+++ b/src/renderer/src/settings/AITab.tsx
@@ -798,10 +798,10 @@ const AITab: React.FC = () => {
       aria-label={label}
     >
       <span
-        className={`absolute left-0.5 top-1/2 h-4 w-4 -translate-y-1/2 rounded-full border shadow-sm transition-transform ${
+        className={`absolute top-1/2 h-4 w-4 -translate-y-1/2 rounded-full border shadow-sm transition-all ${
           enabled
-            ? 'translate-x-[18px] bg-[var(--bg-overlay-strong)] border-[var(--ui-segment-border)]'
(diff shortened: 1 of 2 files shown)


shobhit99:
Hey @monotykamary which version of supercmd are you on right now? i can't find the first toggle button for me on my version
image

monotykamary:
Oh interesting, I see it on v1.0.22

image

image

---

but now that I realized, it's to do with the font-size (I'm using `small`). I'll likely reweave this PR to make the styling calc more consistent.

monotykamary:
Oh I realized the calc is pretty consistent across the font-sizes already. Should be golden @shobhit99. Let me update the description a bit.

PR #390, opened 2026-05-15
Fix image clipboard entries missing from Image tab

### LINKED ISSUE #389: Fix image clipboard entries missing from Clipboard History
## Problem
Image clipboard items can fail to appear in Clipboard History (All and Image tabs).

## Root cause
The clipboard monitor only fingerprinted exact macOS UTI names such as public.png/public.tiff. Some image copies are exposed by Electron as MIME formats like image/png or image/jpeg, so they were ignored before a history item could be saved as type image.

## Expected behavior
Copied screenshots and browser images should be captured as image history entries and show in the Clipboard history.

denysbutenko:
## Summary
- recognize image clipboard formats exposed as MIME types as well as macOS UTIs
- keep raw-buffer fingerprinting for common formats to preserve the clipboard CPU optimization
- fall back to Electron native image decoding only for unknown image-like formats

Fixes #389

## Validation
- npm run build:main
- manually verified copied image appears in Clipboard History > Image tab in local dev

### DIFF
--- a/src/main/clipboard-manager.ts
+++ b/src/main/clipboard-manager.ts
@@ -537,56 +537,100 @@ function isImageFilePath(filePath: string): boolean {
 }
 
 /**
- * Compute a cheap fingerprint for the image currently on the clipboard WITHOUT
- * calling toPNG() / decoding pixels.  Strategy (in priority order):
+ * Compute a cheap fingerprint for the image currently on the clipboard without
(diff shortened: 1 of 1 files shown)


denysbutenko:
Superseded by a guidelines-compliant branch/commit in a new PR using the required fix/... branch naming and Conventional Commit format.

PR #395, opened 2026-05-15
fix(clipboard): add universal nativeImage fallback for unrecognized clipboard format names

monotykamary:
## Problem

When copying images to the clipboard, they do not appear in SuperCmd's clipboard history. Copying any image (screenshot, browser image, etc.) and opening the clipboard panel shows nothing.

## Root Cause

Commit `1349d14` refactored `getClipboardImageFingerprint()` to use format-name matching (`looksLikeClipboardImageFormat`) before attempting to read the image. The native `clipboard.readImage()` fallback was placed **inside** the `if (formats.some(looksLikeClipboardImageFormat))` guard. 

If Electron exposes format names that don't match our list (e.g. dynamic UTIs like `dyn.ah62d4rv4gu8yc6c`, legacy NSPasteboard types, or any format name not in the allowlist), the function returns `{ fingerprint: "" }` immediately — `clipboard.readImage()` is **never called**, and the image is silently skipped.

## Fix

Restructured `getClipboardImageFingerprint()` with two fallback paths:

1. **Fast path** (inside `if (formats.some(looksLikeClipboardImageFormat))`) — tries raw format buffers (GIF → PNG → JPEG → WebP → HEIC → TIFF) for fingerprinting without `toPNG()`. Falls back to `clipboard.readImage()` for any matched-but-unreadable format.

2. **Universal fallback** (outside the format-name guard) — tries `clipboard.readImage()` regardless of format names. Catches dynamic UTIs, legacy types, or any unrecognized format name.

The CPU optimization is preserved: `toPNG()` only runs once per new image (hash comparison prevents re-processing), and `clipboard.readImage()` on an empty clipboard returns immediately via `isEmpty()`.

## Changes

- `src/main/clipboard-manager.ts` — `getClipboardImageFingerprint()`: moved the format-name guard to only gate the fast buffer path; added a universal `clipboard.readImage()` fallback that runs unconditionally.

### DIFF
--- a/src/main/clipboard-manager.ts
+++ b/src/main/clipboard-manager.ts
@@ -586,45 +586,59 @@ function getClipboardImageFingerprint(): {
 } {
   try {
     const formats = clipboard.availableFormats();
-    if (!formats.some(looksLikeClipboardImageFormat)) return { fingerprint: '' };
 
-    const gifBuf = readClipboardBufferForFormats(formats, ['com.compuserve.gif', 'image/gif']);
-    if (gifBuf && gifBuf.length > 4 &&
-        gifBuf[0] === 0x47 && gifBuf[1] === 0x49 && gifBuf[2] === 0x46) {
-      return {
-        fingerprint: buildImageFingerprint('gif', gifBuf),
-        rawGifData: gifBuf,
-      };
+    // Fast path: try to read raw format bytes for fingerprinting.
+    // This avoids calling toPNG() on every poll tick for common formats.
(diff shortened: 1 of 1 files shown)


shobhit99:
Hey @monotykamary this already got fixed 😄 it was some regression in last release.
https://github.com/SuperCmdLabs/SuperCmd/pull/391
https://github.com/SuperCmdLabs/SuperCmd/pull/394

monotykamary:
oh totally forgot to see those, just had this on my fork for a while after testing

carry on carry on 🫡

PR #402, opened 2026-05-16
fix: resolve active Chromium profile for browser history import

### LINKED ISSUE #401: Browser history import doesn't detect non-default Chrome profiles
## Description
Browser history import fails for Chromium users who use non-default profiles (e.g., 'Profile 7' instead of 'Default').

## Root Cause
In `src/main/browser-search-history.ts`, all Chromium-family browsers hardcode the history path to `Default/History`:
```typescript
{ id: 'chrome', name: 'Google Chrome', dbPath: path.join(home, 'Library/Application Support/Google/Chrome/Default/History') },
```

If the user's Chrome profile is 'Profile 7', the actual history file is at `Profile 7/History` — but the code never looks there. The `fileExists()` check returns false, so the browser shows as unavailable in the import dropdown.

## Impact
Affects all Chromium browsers (Chrome, Arc, Brave, Edge, Vivaldi) for any user with a non-default profile. This is a common setup for users who separate work/personal browsing.

## Note
Firefox handles this correctly — it scans the `Profiles/` directory. The same pattern should be applied to Chromium browsers.

The `Local State` file in each browser's data directory contains a `profile.last_used` field indicating the active profile, which can be used for detection.

vingeraycn:
## Problem
All Chromium-family browsers hardcode `Default/History` as the history file path in `listImportableBrowsers()`. Users with non-default Chrome profiles (e.g. 'Profile 7') cannot import their browser history — the browser shows as unavailable in the import dropdown.

Fixes #401

## Solution
Added `resolveChromiumHistoryPath(rootDir)` which dynamically resolves the correct profile directory:

1. **Local State** — Reads `{rootDir}/Local State` JSON to find `profile.last_used` (the active profile)
2. **Default fallback** — If Local State is missing or the profile has no History file, falls back to `Default/History`
3. **Profile scan** — If Default doesn't exist either, scans all `Profile *` and `Default` subdirectories, returning the first with a History file
4. **Last resort** — Returns `Default/History` path even if it doesn't exist (caller checks existence)

## Changes
- Added `resolveChromiumHistoryPath()` helper with the 4-tier resolution algorithm
- Changed Chromium browser definitions from hardcoded `dbPath` to `rootDir`
- Display name includes profile name when not 'Default' (e.g., 'Google Chrome (Profile 7)')
- **No changes to `ImportableBrowser` interface** — fully backward compatible
- **No changes to `importFromBrowser()` signature** — existing callers unaffected

## Testing
- TypeScript compilation passes (tsc --noEmit)
- Main build passes (npm run build:main)
- Manually verified that Chrome with 'Profile 7' is now correctly detected

### DIFF
--- a/src/main/browser-search-history.ts
+++ b/src/main/browser-search-history.ts
@@ -373,20 +373,87 @@ function dirExists(p: string): boolean {
   }
 }
 
+/**
+ * Resolve the path to a Chromium browser's History SQLite file.
+ *
+ * Algorithm:
+ *  1. Read `Local State` JSON from the browser root dir; use `profile.last_used`
+ *     to pick the active profile directory.
+ *  2. If that profile has a History file → use it.
+ *  3. If Local State is missing or lacks the field → fall back to `Default/History`.
+ *  4. If Default doesn't exist either → scan subdirectories matching `Profile *`
+ *     and `Default`, returning the first one that contains a History file.
+ *  5. As a last resort, return the `Default/History` path (caller checks existence).
+ */
+function resolveChromiumHistoryPath(rootDir: string): { dbPath: string; profileName: string } {
+  // 1. Try Local State for the last-used profile.
+  const localStatePath = path.join(rootDir, 'Local State');
+  try {
+    const raw = fs.readFileSync(localStatePath, 'utf-8');
+    const state = JSON.parse(raw);
+    const lastUsed: string | undefined = state?.profile?.last_used;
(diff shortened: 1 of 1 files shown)


shobhit99:
@vingeraycn Thanks for the PR. this was actually added in another PR. closing this one.

vingeraycn:
> @vingeraycn Thanks for the PR. this was actually added in another PR. closing this one.

Thanks for your work. @shobhit99

PR #408, opened 2026-05-17
fix: close Clipboard Manager in one Escape when opened via shortcut (#407)

### LINKED ISSUE #407: Clipboard Manager requires two Escape presses to close when opened via shortcut
## Summary

When the Clipboard Manager is opened directly via its global shortcut, one Escape press does not close it. The first Escape drops back to the main launcher list, and a second Escape is needed to dismiss the window. When the Clipboard Manager is opened by shortcut (not navigated to from the launcher), a single Escape should close the window.

## Steps to reproduce

1. Use the global shortcut bound to "Clipboard Manager" (`system-clipboard-manager`).
2. Press the shortcut — Clipboard Manager opens directly.
3. Press Escape once.

**Actual:** The view drops back to the main launcher search list and the window stays open. A second Escape is needed to hide the window.

**Expected:** Since it was opened via shortcut, one Escape closes/hides the window.

## Root cause

The Clipboard Manager Escape handler calls `onClose()` (`src/renderer/src/ClipboardManager.tsx:485-488`). The `onClose` passed from `App.tsx` only runs `setShowClipboardManager(false)` and returns to the launcher list (`src/renderer/src/App.tsx:4041-4046`); it does not hide the window. The window only hides on the next Escape, handled by the launcher's own Escape logic, so it takes two presses.

That behavior is correct when the user navigated into Clipboard Manager from the launcher list (Escape → back to the list). It is wrong when Clipboard Manager was opened directly via shortcut, where there is no list to return to.

Note: if the Cmd+K Actions overlay is open, its own Escape closes that overlay first (`ClipboardManager.tsx:440-445`). That layer is expected and out of scope here.

## Proposed fix

Distinguish "opened via shortcut" from "navigated from launcher", and make Escape close the window in one press for the shortcut case.

The signal already exists: when launched via the routed global shortcut, the `window-shown` payload carries `systemCommandId === 'system-clipboard-manager'` (read in `App.tsx` as `routedSystemCommandId`, around `App.tsx:832`). When opened from the launcher list via `openClipboardManager()` (`src/renderer/src/hooks/useAppViewManager.ts:181-184`), no such payload is set.

- Thread an `openedViaShortcut` prop into `ClipboardManager`, derived from whether it was opened from the routed `systemCommandId` payload vs. launcher navigation.
- In the `onClose` handler in `App.tsx`, when `openedViaShortcut` is true, hide/close the window directly (e.g. `closeMainWindow` / window hide) instead of only `setShowClipboardManager(false)`.
- Keep the current behavior (Escape → back to launcher list) when opened from the launcher.

## Acceptance criteria

- [ ] Opened via shortcut: one Escape closes/hides the window.
- [ ] Opened from the launcher list: Escape returns to the launcher list (unchanged); a second Escape closes the window (unchanged).
- [ ] Cmd+K Actions overlay still consumes its own Escape first in both cases.

## Affected files

- `src/renderer/src/ClipboardManager.tsx`: Escape handler (~485-488)
- `src/renderer/src/App.tsx`: `onClose` (~4041-4046), `routedSystemCommandId` (~832)
- `src/renderer/src/hooks/useAppViewManager.ts`: `openClipboardManager` (~181-184)

denysbutenko:
## What

Closes #407. Opened straight from its global shortcut, Clipboard Manager needed two Escape presses to dismiss. The first press dropped back to the launcher list; the second hid the window. When there's no list to go back to, one Escape should close it.

## Why it happened

`ClipboardManager`'s `onClose` in `App.tsx` always ran `setShowClipboardManager(false)`, which returns to the launcher list no matter how the manager was opened. That's right when you navigated in from the list, wrong when you opened it from a shortcut.

## The fix

A new `clipboardManagerOpenedViaShortcut` flag in `useAppViewManager` records how it was opened:

- `onWindowShown` routed payload (`systemCommandId === 'system-clipboard-manager'`): shortcut
- `runLocalSystemCommand` with `fromMainEvent`: shortcut (the 180 ms `run-system-command` fallback `main.ts` fires for window-shown-routed commands)
- launcher-list selection via `handleCommandExecute`, no options: not a shortcut

Worth calling out: a shortcut hits `openClipboardManager` twice, once from the `window-shown` payload and again 180 ms later from the fallback. Both shortcut paths pass `true`, so the late fallback can't flip the flag back to `false`. When the flag is set, `onClose` hides the window directly; otherwise the old return-to-list behavior is untouched. The Cmd+K actions overlay still consumes its own Escape first.

## Testing

No renderer/IPC test harness in the repo, so this was manual.

- Shortcut path: one Escape hides the window. Confirmed at the OS level: `mainWindow.isVisible()` was `false` right after hide and still `false` 150 ms later, across several open/close cycles.
- Launcher-list path: Escape goes back to the list, second Escape hides. Unchanged.
- Cmd+K overlay closes on its own Escape first in both cases.
- `tsc` on the touched files adds no new type errors; `build:main` and `build:renderer` both pass.

All acceptance criteria in #407 check out.

### DIFF
--- a/src/renderer/src/App.tsx
+++ b/src/renderer/src/App.tsx
@@ -390,12 +390,12 @@ const App: React.FC = () => {
   const homeDir = String((window.electron as any).homeDir || '');
   const {
     extensionView, extensionPreferenceSetup, scriptCommandSetup, scriptCommandOutput,
-    showClipboardManager, showSnippetManager, showNotesSearch, showCanvasSearch, showQuickLinkManager, showFileSearch, showCursorPrompt,
+    showClipboardManager, clipboardManagerOpenedViaShortcut, showSnippetManager, showNotesSearch, showCanvasSearch, showQuickLinkManager, showFileSearch, showCursorPrompt,
     showWhisper, showSpeak, showCamera, showSchedule, showWindowManager, showAppUninstall, showWhisperOnboarding, showWhisperHint, showOnboarding, aiMode,
     openOnboarding, openWhisper, openClipboardManager,
     openSnippetManager, openNotesSearch, openCanvasSearch, openQuickLinkManager, openFileSearch, openCursorPrompt, openSpeak, openCamera, openSchedule, openWindowManager, openAppUninstall,
     setExtensionView, setExtensionPreferenceSetup, setScriptCommandSetup, setScriptCommandOutput,
-    setShowClipboardManager, setShowSnippetManager, setShowNotesSearch, setShowCanvasSearch, setShowQuickLinkManager, setShowFileSearch, setShowCursorPrompt,
+    setShowClipboardManager, setClipboardManagerOpenedViaShortcut, setShowSnippetManager, setShowNotesSearch, setShowCanvasSearch, setShowQuickLinkManager, setShowFileSearch, setShowCursorPrompt,
     setShowWhisper, setShowSpeak, setShowCamera, setShowSchedule, setShowWindowManager, setShowAppUninstall, setShowWhisperOnboarding, setShowWhisperHint,
     setShowOnboarding, setAiMode,
   } = useAppViewManager();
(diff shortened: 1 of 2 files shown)


denysbutenko:
Hi @shobhit99 can you review this PR?

shobhit99:
tested, works!

PR #446, opened 2026-05-26
fix: prevent settings window flash when whisper overlay opens

monotykamary:
## Problem

When the settings window (or canvas/extension-store window) was previously opened, pressing the whisper hotkey causes a brief **flash of the settings window** before the whisper overlay appears.

## Root Cause

When `openSettingsWindow()` is called, it switches the SuperCmd app to macOS **"regular" activation mode** (`setMacActivationPolicy("regular")`), which makes the app appear in the Dock and Cmd+Tab.

When the whisper hotkey is pressed, the renderer calls `window.open()` to create the detached whisper popup. Electron intercepts this via `setWindowOpenHandler` and creates a new `BrowserWindow` with `show: true`. On macOS, `show: true` calls `makeKeyAndOrderFront:` which **activates the application**, raising all existing app windows to the foreground — including the settings window. The user sees the settings window flash into view before the `alwaysOnTop: true` whisper popup renders on top.

## Fix

Change the whisper popup creation to use `show: false` and then call `showInactive()` after setting up `setIgnoreMouseEvents(true)` in the `did-create-window` handler.

`showInactive()` maps to `NSWindow orderFrontRegardless:` on macOS, which orders the window in front **without making it key or activating the application**. This prevents macOS from raising the settings window when the whisper popup appears.

This approach:
- Adds **zero delay** — no window hiding/showing, no activation policy switches
- Does **not restore** the settings window after whisper closes (the user did not ask for it)
- Is **minimally invasive** — only 9 lines changed, no new state variables

### Changes

- `src/main/main.ts`:
  - In `setWindowOpenHandler`: changed `show: true` to `show: detachedPopupName !== DETACHED_WHISPER_WINDOW_NAME` so the whisper popup is created hidden
  - In `did-create-window`: added `childWindow.showInactive()` after `setIgnoreMouseEvents(true)` for the whisper popup, showing it without activating the app

## Testing

1. Open SuperCmd settings (Cmd+, or from menu)
2. Switch to another app (e.g. TextEdit)
3. Press the whisper hotkey
4. **Before fix**: settings window flashes briefly before the whisper overlay appears
5. **After fix**: whisper overlay appears cleanly with no settings window flash and no added delay

### DIFF
--- a/browser-extension/background.js
+++ b/browser-extension/background.js
@@ -1,36 +1,75 @@
-const SUPERCMD_ENDPOINT = 'http://127.0.0.1:17373/browser-tabs/snapshot';
-const SUPERCMD_COMMANDS_ENDPOINT = 'http://127.0.0.1:17373/browser-tabs/commands';
-const SUPERCMD_COMMAND_RESULT_ENDPOINT = 'http://127.0.0.1:17373/browser-tabs/command-result';
+const SUPERCMD_BASE_URL = 'http://127.0.0.1:17373';
+const SNAPSHOT_ENDPOINT = `${SUPERCMD_BASE_URL}/browser-tabs/snapshot`;
+const HELLO_ENDPOINT = `${SUPERCMD_BASE_URL}/browser-tabs/hello`;
+const COMMANDS_ENDPOINT = `${SUPERCMD_BASE_URL}/browser-tabs/commands`;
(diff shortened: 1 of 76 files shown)


shobhit99:
@monotykamary can you resolve the conflicts

monotykamary:
oh just noticed, gimme a hot min 🚐

monotykamary:
@shobhit99 all good 🫡 
it kind of merged seamlessly actually, not sure why there were conflict markers 🤔

PR #468, opened 2026-06-13
fix(extension-runner): symlink assetsPath to avoid spaces breaking execSync in extensions

ajchemist:
## Summary

- Extensions like `sips` (Image Modification) call `chmod +x ${assetsPath}/...` via `execSync` without quoting the path
- `assetsPath` includes `Application Support` which contains a space, causing the shell to split the argument and fail with `Command failed: chmod +x /Users/.../Application`
- Fix: when `assetsPath` contains spaces, create a space-free symlink at `/tmp/supercmd-assets/<extname>` pointing to the real assets directory and expose that as `assetsPath`

## Root Cause

```
// Extension code (cannot be modified)
execSync(`chmod +x ${environment.assetsPath}/webp/arm/cwebp`);
// expands to (unquoted, space in path):
// chmod +x /Users/.../Application Support/SuperCmd/extensions/sips/assets/webp/arm/cwebp
// shell sees "Application" and "Support/..." as separate args → ENOENT
```

## Test Plan

Tested locally via `npm run dev`:

- [x] Open SuperCmd → run **Convert Images** command from Image Modification extension
- [x] Select an image and choose **WEBP** format
- [x] Confirmed conversion succeeds without "Command failed: chmod +x" error

### DIFF
--- a/src/main/extension-runner.ts
+++ b/src/main/extension-runner.ts
@@ -1304,7 +1304,23 @@ export async function getExtensionBundle(
   } catch {}
 
   // Compute paths
-  const assetsPath = path.join(extPath, 'assets');
+  const rawAssetsPath = path.join(extPath, 'assets');
+  // Some extensions run `chmod +x ${assetsPath}/...` via execSync without quoting the path.
+  // If assetsPath contains spaces (e.g. "Application Support"), the shell splits on them and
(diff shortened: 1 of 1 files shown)


shobhit99:
Thanks for fixing this, fix is working

PR #476, opened 2026-06-24
fix: surface individual settings like "Color Filters" in search (#471)

### LINKED ISSUE #471: Unable to find "color filters"
It can easily be found in spotlight but SuperCmd is facing issues.

Image

shobhit99:
## Summary

Fixes #471 — settings findable in Spotlight (like **Color Filters**) couldn't be found in SuperCmd.

System Settings discovery only added the **top-level pane** (e.g. "Accessibility") to the command list. macOS exposes each pane's individual settings in a `.searchTerms` plist inside the settings bundle, and SuperCmd already had a function to parse it — `discoverSettingsSearchTermCommands()` — but **it was never called**. It was effectively dead code.

This PR wires that function into the appex (Ventura+) settings discovery loop, so each pane's individual settings are added to the searchable command list, by name and keyword, just like Spotlight.

**Example:** "Color Filters" lives in the Accessibility bundle's `UniversalAccess.searchTerms` as `AX_DISPLAY_FILTER_ENABLED` → title `"Color filters"`, keywords `enable color filters, color blind, colorblind`. After this change, searching "color filters" returns it, opening to Accessibility settings.

### Changes
- `src/main/commands.ts`: call `discoverSettingsSearchTermCommands()` for each discovered appex pane and append the resulting sub-settings to the command list.
- Bump `COMMANDS_DISK_CACHE_VERSION` 1 → 2 so existing installs refresh their cached command list and pick up the new sub-settings on first launch.

Each sub-item carries the parent pane's icon and a subtitle of the pane name, reuses the pane's open identifier (selecting it opens the parent pane), and is excluded from the generic-icon dedup pass because it has a subtitle.

## Test Coverage

The discovery path reads live macOS system bundles and the helper is internal (not exported), so it isn't covered by the existing pure-function test suite. Verified manually:
- Simulated the extraction against the real `AccessibilitySettingsExtension.appex/.../UniversalAccess.searchTerms` — "Color filters" is produced with the expected title and keywords.
- `npm test` — 27/27 pass (root-search ranking suite).
- `npm run check:i18n` — all locales OK (no user-facing strings changed; sub-setting titles come from macOS).
- `npm run build:main` — clean.

## Test plan
- [x] Existing tests pass (27/27)
- [x] i18n parity check passes
- [x] Main process builds
- [ ] Manual: search "color filters" in SuperCmd → result appears and opens Accessibility settings

### DIFF
--- a/scripts/test-root-search-ranking.mjs
+++ b/scripts/test-root-search-ranking.mjs
@@ -70,7 +70,7 @@ const {
   scoreRootSearchCandidate,
   scoreRootSearchFields,
 } = ranking;
-const { assembleRootSearchSections } = sections;
+const { assembleRootSearchSections, isRootResultPromotionCandidate } = sections;
 
 const now = Date.UTC(2026, 4, 17);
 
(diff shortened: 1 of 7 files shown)


shobhit99:
## Codex review (gpt, reasoning=high) — GATE: PASS

One [P2] finding, now fixed in `7bfecbf1`:

> **Avoid surfacing raw searchTerms keys** — `discoverSettingsSearchTermCommands` added each top-level `.searchTerms` key as its own visible command. Those keys are internal identifiers (e.g. `AX_ALT_MOUSE_ENABLE_SOUNDS`), so wiring the function in flooded the launcher with junk commands.

Verified and fixed:
- Across all 47 system panes with `.searchTerms` files, only **2** keys lack `localizableStrings` rows — every other key carries the real human label in `localizableStrings[].title`.
- Dropped the section-key-as-title branch and its identifier-derived keywords; now only row titles (plus their curated `index` keywords) are surfaced.
- Accessibility sub-items: **679 → 342** clean entries, **0** junk identifiers, "Color Filters" still resolves.

No other findings. `npm test` 27/27, `build:main` clean.

PR #481, opened 2026-07-03
perf(search): remove root ranking diagnostics

JustYannicc:
## What changed
- Removed the temporary `SC-RANK2` diagnostic effect from the root search command model.
- Removed the root search ranking startup/liveness `console.info` stamp.

## Why
- The diagnostic effect ran as the root search query/ranked candidates changed, scanning candidates and sending `window.electron.whisperDebugLog` IPC from the per-keystroke ranking path.
- The startup stamp was only for verifying an earlier ranking fix and should not run in production startup paths.

Before evidence:
- `node scripts/test-root-search-ranking.mjs`: passed after hydrating dependencies; emitted the startup liveness log.
- Debug-path measurement: `{"phase":"before","scRank2Tags":3,"rootSearchWhisperDebugLogCallsites":2,"tempDiagnosticEffects":1,"startupLivenessLogs":1}`.

After evidence:
- Debug-path measurement: `{"phase":"after","scRank2Tags":0,"rootSearchWhisperDebugLogCallsites":0,"tempDiagnosticEffects":0,"startupLivenessLogs":0}`.

## Compatibility impact
- Ranking scores, result ordering, section assembly, and user-facing behavior are unchanged.
- No user-facing strings were added, renamed, or removed, so no locale files changed.
- Debug-only `SC-RANK2` IPC/log output is no longer available from the root search hot path.

## How tested
- `node scripts/test-root-search-ranking.mjs` before the fix: pass.
- `node scripts/test-root-search-ranking.mjs` after the fix: pass.
- Codex LSP diagnostics:
  - `src/renderer/src/utils/root-search-ranking.ts`: no diagnostics.
  - `src/renderer/src/hooks/useLauncherCommandModel.ts`: reports existing `BrowserSearchSource` typing diagnostics around profile browser IDs outside this diff.
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit --pretty false`: fails on existing repo-wide TypeScript diagnostics, including the hook profile typing diagnostics above plus unrelated files such as `App.tsx`, `CameraExtension.tsx`, `LiquidGlassSurface.tsx`, and Raycast runtime modules.

### DIFF
--- a/src/renderer/src/hooks/useLauncherCommandModel.ts
+++ b/src/renderer/src/hooks/useLauncherCommandModel.ts
@@ -857,41 +857,6 @@ export function useLauncherCommandModel({
     queryFileSectionCommands,
   } = rootSearchSectionAssembly;
 
-  // TEMP DIAGNOSTIC (SC-RANK2): pinpoint whether the internal>browser comparator
-  // is actually running, and whether Search Notes / Create Note are candidates.
-  // Remove once the override bug is confirmed fixed.
-  useEffect(() => {
-    if (!hasSearchQuery) return;
-    try {
-      const idOf = (c: any) => String(c?.command?.id || c?.stableKey || '');
-      const sn = commandCandidates.find((c) => idOf(c).includes('search-notes'));
-      const cn = commandCandidates.find((c) => idOf(c).includes('create-note'));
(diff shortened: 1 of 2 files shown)


JustYannicc:

PR #482, opened 2026-07-03
perf(cache): sweep expired withCache entries

JustYannicc:
## Summary
- Sweep expired `withCache` entries before inserting a fresh result when `maxAge` is configured.
- Delete an expired cache hit before recomputing the value for that key.
- Add focused regression coverage for expired unique keys, expired-hit recompute, and non-expiring cache behavior.

## Root Cause
`withCache` only checked expiry when a key was reused. Long-lived extension sessions that generated many one-off keys could leave expired entries in the internal `Map` forever because no later lookup touched those keys.

## Before / After Evidence
- Before the fix, the new focused test failed with `[withCache baseline] retained expired unique entries: 1001` after 1,000 expired unique keys plus one latest key.
- After the fix, the same scenario passes with `[withCache metric] retained entries after unique expired inserts: 1`.

## Validation
- `node --test scripts/test-with-cache-expiry.mjs`
- `pnpm test`
- `pnpm run check:i18n`
- `pnpm run build:renderer`
- Codex LSP diagnostics on `src/renderer/src/raycast-api/platform-runtime.ts`: no diagnostics
- Codex LSP references for `withCache`: export plus re-export sites only

## Notes / Risks
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit` still fails on existing renderer type errors unrelated to this change, including `App.tsx`, `CameraExtension.tsx`, `LiquidGlassSurface.tsx`, launcher hooks, and other Raycast runtime type issues.
- Sweep cost is O(cache size) only on cache misses/inserts with `maxAge` configured. Non-expiring caches do not run the sweep.

### DIFF
--- a/scripts/test-with-cache-expiry.mjs
+++ b/scripts/test-with-cache-expiry.mjs
@@ -1,0 +1,119 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+function loadPlatformRuntimeWithObservedMaps() {
+  const filePath = path.resolve('src/renderer/src/raycast-api/platform-runtime.ts');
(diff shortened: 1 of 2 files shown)


JustYannicc:

PR #483, opened 2026-07-03
perf(system): gate auto-quit AppleScript probes

JustYannicc:
## Summary
- Gate auto-quit AppleScript probes behind due-candidate checks.
- Cache tracked bundle IDs and tracked music bundle IDs so each poll can prune and decide probe needs without rebuilding sets.
- Add focused auto-quit instrumentation/tests for no-due, due, frontmost, recording, and music gating behavior.

## Metrics
- Before: no tracked app due over 3 ticks = 9 osascript calls (frontmost 3, recording 3, music 3, quit 0).
- After: no tracked app due over 3 ticks = 0 osascript calls.
- Before: one non-music app due = 4 osascript calls (frontmost 1, recording 1, music 1, quit 1).
- After: one non-music app due = 3 osascript calls (frontmost 1, recording 1, music 0, quit 1).

## Validation
- node scripts/test-auto-quit-manager.mjs
- node scripts/test-auto-quit-manager.mjs --metrics
- node --test scripts/test-*.mjs
- pnpm run build:main
- Codex LSP diagnostics on src/main/auto-quit-manager.ts: no diagnostics found

## Risks
- The manager now samples the frontmost app only once at least one tracked app reaches its due time. This is the intended probe reduction, but it means short foreground activations between due checks are not observed until a due candidate exists.

### DIFF
--- a/scripts/test-auto-quit-manager.mjs
+++ b/scripts/test-auto-quit-manager.mjs
@@ -1,0 +1,310 @@
+#!/usr/bin/env node
+
+import assert from 'assert/strict';
+import fs from 'fs';
+import path from 'path';
+import vm from 'vm';
+import { createRequire } from 'module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+const { promisify } = require('util');
+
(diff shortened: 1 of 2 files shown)


JustYannicc:

PR #485, opened 2026-07-03
perf(native): use index queues for AX BFS

JustYannicc:
## What changed

- Replaced `Array.removeFirst()` queue consumption in the native Accessibility BFS loops with monotonic index cursors.
- Added `scripts/bench-ax-bfs-queue.swift` so the queue behavior benchmark is repeatable.

## Why

Accessibility selected-text and caret lookup run on hot paths. `Array.removeFirst()` shifts/copies the remaining queue on every pop, even though these traversals only need FIFO order and already cap depth/element counts.

## Behavior and compatibility

- Preserves BFS search order, focused-child preference, depth limits, and selected/caret lookup fallback order.
- No user-facing strings changed.

## Evidence

- Baseline before editing, synthetic Swift BFS queue benchmark: `removeFirst: 235.47 ms`, `cursor: 37.95 ms`, equal checksum, `6.20x` speedup.
- Repeatable benchmark after change: `swift -O scripts/bench-ax-bfs-queue.swift` -> `removeFirst: 211.29 ms`, `cursor: 36.80 ms`, equal checksum, `5.74x` speedup.
- LSP diagnostics: no errors for `src/native/ax-caret-query.swift`, `src/native/get-selected-text.swift`, and `scripts/bench-ax-bfs-queue.swift`.
- `swiftc -O -o /tmp/supercmd-get-selected-text-check src/native/get-selected-text.swift -framework Foundation -framework ApplicationServices -framework AppKit`.
- `swiftc -O -o /tmp/supercmd-emoji-trigger-monitor-check src/native/emoji-trigger-monitor.swift src/native/ax-caret-query.swift -framework AppKit -framework ApplicationServices`.
- `node --test 'scripts/test-*.mjs'` passed: 24 passed, 1 skipped.
- `node scripts/check-i18n.mjs` exited 0; it reports pre-existing Italian locale gaps unrelated to this native-only change.

## Risk

Low. The traversal queue still appends in the same order and stops at the same caps; the only intentional change is avoiding per-pop array shifting.

### DIFF
--- a/scripts/bench-ax-bfs-queue.swift
+++ b/scripts/bench-ax-bfs-queue.swift
@@ -1,0 +1,106 @@
+import Foundation
+
+private struct Node {
+  let firstChild: Int
+  let childCount: Int
+}
+
+private let depthLimit = 8
+private let branching = 3
+private var nodes: [Node] = []
+private var frontier: [(index: Int, depth: Int)] = [(0, 0)]
+
+nodes.reserveCapacity(9_841)
+nodes.append(Node(firstChild: -1, childCount: 0))
+
+var frontierIndex = 0
+while frontierIndex < frontier.count {
+  let (nodeIndex, depth) = frontier[frontierIndex]
+  frontierIndex += 1
+  guard depth < depthLimit else { continue }
+
+  let firstChild = nodes.count
+  for _ in 0..<branching {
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #486, opened 2026-07-03
perf(menubar): cache visible payload hashes

JustYannicc:
## What changed

- Added a small renderer-side MenuBarExtra visible-payload hash cache.
- MenuBarExtra now refreshes the current action map on every sync, then skips `updateMenuBar` when the serialized visible payload is unchanged.
- Added focused regression coverage for repeated equivalent registrations, visible title/icon/menu changes, and action-map-before-cache ordering.

## Why

Equivalent menu-bar registrations could repeatedly send the same visible payload to the main process. Each IPC update made main rebuild the tray icon/menu even when the native result would be identical.

## Compatibility impact

No extension API or visible UI behavior changes are intended. Changed title, icon, tooltip, menu items, disabled state, and action presence still send a fresh payload. Callback-only action changes keep working because the renderer action map is updated before the unchanged payload is skipped.

## How tested

- Baseline pre-change instrumentation: 3 equivalent visible menu-bar registrations produced 3 renderer `updateMenuBar` sends and 3 action-map updates.
- After regression test: 3 equivalent registrations produce 1 renderer send and still perform 3 action-map updates.
- `node --test scripts/test-menubar-payload-cache.mjs` passes.
- `pnpm run test` passes: 28 passed, 1 skipped.
- Codex LSP diagnostics: no errors in touched TypeScript files.
- `pnpm run build:renderer` passes.
- `pnpm run build:main` passes.
- `node scripts/check-i18n.mjs` completed; it reports pre-existing Italian missing keys unrelated to this no-string change.
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit` still fails on existing unrelated project-wide renderer type errors outside this change.

### DIFF
--- a/scripts/test-menubar-payload-cache.mjs
+++ b/scripts/test-menubar-payload-cache.mjs
@@ -1,0 +1,86 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { importTs } from './lib/ts-import.mjs';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const {
+  createMenuBarVisiblePayloadHashCache,
+  shouldSendMenuBarVisiblePayload,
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #487, opened 2026-07-03
perf(window-management): make AeroSpace workspace moves async

JustYannicc:
## Summary
- Extracted AeroSpace workspace reconciliation into `src/main/aerospace-workspace.ts`.
- Replaced synchronous `execFileSync('aerospace', ...)` calls from the launcher show path with async `execFile`-backed fire-and-forget work.
- Added in-flight coalescing so repeated show/post-show requests produce at most one follow-up reconciliation while a run is active.
- Added regression coverage for coalescing, missing `aerospace` binary behavior, and non-blocking slow-command behavior.

## Why
`showWindow()` called `moveWindowToCurrentAerospaceWorkspace()` before showing/focusing the launcher and again shortly after show. That helper performed multiple synchronous CLI calls on the Electron main process, so slow `aerospace` startup/server responses could directly delay launcher visibility, focus, and menu-bar responsiveness.

## Performance Evidence
Controlled fake `aerospace` command delay: 120 ms per invocation.

Before:
- sync helper duration: 734 ms
- zero-delay timer fired after: 734 ms

After:
- `requestMove()` returned after: 0 ms
- zero-delay timer fired after: 0 ms
- background reconciliation idle after: 364 ms
- command count: 3

## Validation
- `node --test scripts/test-aerospace-workspace.mjs`
- `/opt/homebrew/Cellar/node@22/22.22.3/bin/npm test` (live Electron crash recovery self-skipped because Electron was not launchable here)
- `/opt/homebrew/Cellar/node@22/22.22.3/bin/npm run check:i18n` (exits 0; existing Italian missing keys reported: `settings.ai.whisper.vocabulary`, `settings.extensions.appSearchScope`)
- `/opt/homebrew/Cellar/node@22/22.22.3/bin/npm run build:main`
- Codex LSP diagnostics for `src/main`: 0 errors

## Risks
- AeroSpace movement is now eventual rather than blocking before `mainWindow.show()`. The existing post-show reconciliation remains and coalesces with any in-flight pre-show request to preserve placement as closely as possible without holding the main process.
- Non-ENOENT command failures remain transient and retry on later show requests, matching the previous behavior.

### DIFF
--- a/scripts/test-aerospace-workspace.mjs
+++ b/scripts/test-aerospace-workspace.mjs
@@ -1,0 +1,102 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { setTimeout as delay } from 'node:timers/promises';
+import { fileURLToPath } from 'node:url';
+import { importTs } from './lib/ts-import.mjs';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const { createAerospaceWorkspaceMover } = await importTs(path.join(root, 'src/main/aerospace-workspace.ts'));
+
+test('AeroSpace workspace mover', async (t) => {
+  await t.test('coalesces concurrent move requests into one follow-up run', async () => {
+    const calls = [];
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #488, opened 2026-07-03
perf(speech): replace speak settings polling with listener

JustYannicc:
## Summary
- Replaces the speak manager's 2s settings polling loop with one initial settings load plus the existing `settings-updated` preload event.
- Keeps the configured TTS model/voice and runtime `speakOptions` sync behavior in a small tested helper.
- Cleans up the settings listener on unmount.

## Before / After Evidence
- Baseline instrumentation before product edits found idle `getSettings` calls at `t=0, 2000, 4000, 6000ms`, or 4 calls over a 6.5s idle window.
- After this change, the hook has one initial `getSettings()` call on mount and no settings polling interval. While idle, additional settings reads are 0; live changes arrive through `onSettingsUpdated`.
- This removes roughly 30 idle `getSettings` calls per minute per mounted speak manager, about 1,800 calls per hour in a long session.

## Validation
- `node scripts/test-speak-settings-listener.mjs` passed.
- `node --test 'scripts/test-*.mjs'` passed: 28 passing, 1 skipped Electron-live test.
- `node_modules/.bin/tsc -p tsconfig.main.json` passed.
- `node_modules/.bin/vite build` passed.
- Codex LSP diagnostics on `useSpeakManager.ts` and `speak-settings-sync.ts`: no errors.
- `node scripts/check-i18n.mjs` exited 0; it reports two existing Italian missing keys unrelated to this change. No user-facing strings were added or changed.

## Notes
- `npm` is intentionally disabled on this machine and its wrapper attempted a pnpm install that is blocked by pnpm's ignored-build policy, so validation used the underlying script commands directly from `node_modules`.

### DIFF
--- a/scripts/test-speak-settings-listener.mjs
+++ b/scripts/test-speak-settings-listener.mjs
@@ -1,0 +1,131 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+const hookSource = fs.readFileSync('src/renderer/src/hooks/useSpeakManager.ts', 'utf8');
+
+function loadTsModule(filePath) {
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #489, opened 2026-07-03
perf(files): optimize file search delete batches

JustYannicc:
## Summary

- Collapse nested watcher delete paths before tombstoning the file-search index.
- Replace the entries x delete-prefix scan with direct-id tombstones plus descendant prefix/ancestor-set matching.
- Add a focused synthetic benchmark and regression coverage for file deletes, directory deletes, nested delete collapse, and non-matching prefix-like paths.

## Benchmark

Command: `node scripts/benchmark-file-search-delete-batch.mjs`

Synthetic scenario: 68,162 index entries, 4,041 nested delete paths, 8,041 tombstoned entries.

- Before: 13,295.52 ms
- After: 21.42 ms final run (observed after runs: 16.41 ms, 15.85 ms, 25.38 ms, 21.42 ms)
- Improvement: about 621x faster on the final run

## Validation

- `node --test scripts/test-file-search-delete-batch.mjs`
- `node --test 'scripts/test-*.mjs'` (passes; live Electron test skipped because Electron is not launchable here)
- Codex LSP diagnostics on `src/main/file-search-index.ts`: no errors
- `pnpm run build:main`
- `pnpm run check:i18n` exited 0; it still reports pre-existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`, and this PR does not touch user-facing strings

## Risks

- Delete matching remains string/path based and preserves the previous case-sensitive behavior.
- Tombstoned entries remain in `pathToEntryId` as before, so re-added paths can clear the deleted flag through `indexEntry`.

### DIFF
--- a/scripts/benchmark-file-search-delete-batch.mjs
+++ b/scripts/benchmark-file-search-delete-batch.mjs
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+
+import { createRequire } from 'node:module';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import vm from 'node:vm';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+function loadFileSearchIndexInternals() {
+  const filePath = path.resolve('src/main/file-search-index.ts');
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #491, opened 2026-07-03
perf(audio): avoid idle buffer conversion

JustYannicc:
## Summary
- skip `AVAudioConverter` work on warm idle audio taps unless a meter update is due
- compute idle meter levels directly from the input buffer, while keeping recording conversion/output unchanged
- keep warmup and command behavior unchanged; conversion still happens for recording buffers and recording meters

## Evidence
Focused temporary instrumentation around `processBuffer` counted converted, recorded, and metered buffers over matching 1.2s windows:

| Scenario | Before | After |
| --- | ---: | ---: |
| Warm idle conversions | 12 | 0 |
| Warm idle recorded buffers | 0 | 0 |
| Warm idle meter updates | 7 | 7 |
| Recording conversion delta | 12 | 12 |
| Recording append delta | 12 | 12 |
| Recording meter delta | 8 | 8 |

This removes idle converter work while preserving recording conversion and meter cadence in the measured run.

## Checks
- `mcp__lsp.diagnostics` on `src/native/audio-capturer.swift`: no errors
- `swiftc -O -o /tmp/supercmd-audio-capturer-final src/native/audio-capturer.swift -framework AVFoundation -framework Foundation`: passed
- `git diff --check`: passed
- `node scripts/check-i18n.mjs`: completed; existing `it` locale missing keys were reported and are unrelated to this native-only change
- `npm run build:native`: blocked by local policy; npm is disabled on this Mac, pnpm attempted install, then failed with `ERR_PNPM_IGNORED_BUILDS`

## Risks
- Idle meter levels are now calculated from the input buffer instead of the converted 16 kHz mono buffer. The code reads channel 0 with the same RMS/peak scaling, matching the converter's channel map for typical mic formats.

### DIFF
--- a/src/native/audio-capturer.swift
+++ b/src/native/audio-capturer.swift
@@ -279,6 +279,16 @@ class AudioCapturer {
   // MARK: Buffer processing
 
   private func processBuffer(_ buffer: AVAudioPCMBuffer) {
+    let now = Date()
+    let shouldUpdateMeter = now.timeIntervalSince(lastMeterAt) >= meterInterval
+
+    guard isRecording || shouldUpdateMeter else { return }
+
+    if !isRecording {
+      updateMeter(fromInputBuffer: buffer, now: now)
+      return
+    }
+
     guard let converted = convertBuffer(buffer),
           converted.frameLength > 0,
           let samples = converted.floatChannelData?[0]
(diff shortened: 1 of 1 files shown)


JustYannicc:

PR #493, opened 2026-07-03
perf(frecency): memoize sorting defaults

JustYannicc:
## Summary
- Hoists the default `useFrecencySorting` key resolver so stable rerenders do not invalidate `useMemo`.
- Captures one frecency scoring timestamp per sort recompute instead of calling `Date.now()` for each score.
- Adds a node test harness for stable rerenders, timestamp count, ordering, visit tracking, and reset behavior.

## Evidence
- Baseline before the production change: `node --test scripts/test-use-frecency-sorting.mjs` failed with stable rerender comparisons increasing `4 -> 8`, and one sort reading `Date.now()` `8` times instead of `1`.
- After the change: `node --test scripts/test-use-frecency-sorting.mjs` passes all 3 tests.

## Validation
- `npm test` passes: 27 passed, 1 skipped.
- `npm run check:i18n` exits 0; it still reports pre-existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`.
- `npm run build:main` passes.
- `npm run build:renderer` passes.
- LSP diagnostics for the edited hook report no diagnostics.

## Risks
- The default key now stringifies `id` values explicitly; object property access already coerced those keys for persisted records, so existing storage keys should continue to match.

### DIFF
--- a/scripts/test-use-frecency-sorting.mjs
+++ b/scripts/test-use-frecency-sorting.mjs
@@ -1,0 +1,244 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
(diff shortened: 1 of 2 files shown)


JustYannicc:

PR #496, opened 2026-07-03
test: add bundled TypeScript import helper

JustYannicc:
## Summary
- Replace the dependency-free-only `scripts/lib/ts-import.mjs` transform with an esbuild bundle-backed TS/TSX import helper.
- Add safe default stubs for Electron, React/JSX, common renderer globals, and static assets so selected production modules can be imported in Node tests.
- Convert the shallow browser-search history export grep into an executable import test that loads `src/main/browser-search-history.ts` and exercises `resolveInput()` through its relative resolver dependency.

## Baseline
`scripts/test-browser-search-lag-fix.mjs` previously read source text and grepped for browser-search history names. That proved text existed, but it could not prove the module loaded, that relative imports resolved, or that exported behavior still worked.

## Validation
- Codex LSP diagnostics: no errors for `scripts/lib/ts-import.mjs` and `scripts/test-browser-search-lag-fix.mjs`.
- `node --test scripts/test-browser-search-lag-fix.mjs`: 11 pass, 0 fail.
- TSX smoke check: `importTs(src/renderer/src/icons/Calendar.tsx)` returned a default function.
- `npm test`: 25 pass, 0 fail, 1 skipped (`Renderer crash recovery (live Electron)` skipped because Electron is not launchable here).

## Risks
- The helper intentionally stubs host APIs; tests that need real Electron or browser behavior should still use dedicated live harnesses.
- Default stubs are broad enough for import-time tests, but module-specific behavior tests may need custom stubs via `importTs(..., { stubs })`.

### DIFF
--- a/scripts/lib/ts-import.mjs
+++ b/scripts/lib/ts-import.mjs
@@ -1,16 +1,293 @@
-// Import a self-contained TypeScript module from a test by transpiling it with
-// esbuild on the fly. This lets the recovery tests run the *real* production
-// source (renderer-recovery.ts, reload-budget.ts) instead of grepping it.
-//
-// Only works for modules with no relative imports of their own — the recovery
-// helpers are deliberately dependency-free for exactly this reason.
+// Import a TypeScript/TSX module from a test by bundling it with esbuild on the
+// fly. Bundling lets tests execute selected production modules that have local
+// imports instead of falling back to grep-only assertions.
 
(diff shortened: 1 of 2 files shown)


JustYannicc:
Superseded by #532 (fix(validation): restore project checks), which consolidates this validation/build change with the related source PRs. Closing this PR in favor of the upstream replacement.

PR #497, opened 2026-07-03
perf(notes): coalesce autosave persistence

JustYannicc:
## Summary
- Replace synchronous Notes JSON writes on each mutation with a debounced async atomic write queue.
- Keep `notesCache` as the authoritative in-memory state for immediate reads while persistence drains in the background.
- Add `flushNotesToDisk()` / `hasPendingNotesSave()` and defer Electron quit only when a Notes save is pending or in-flight.
- Add a focused Notes persistence regression harness that counts physical Notes JSON writes and verifies shutdown flush behavior.

## Before / After Evidence
- Baseline before production edits: 20 repeated autosaves produced 20 synchronous writes, `loopMs=23.77`, `syncWriteMs=23.11`.
- After: 20 repeated autosaves produce 1 async atomic write, 0 sync writes, latest observed full-suite metric `loopMs=2.62`, `syncWriteMs=0.00`.

## Validation
- `node --test scripts/test-notes-async-saves.mjs`
- `node --test scripts/test-*.mjs` (26 pass, 1 live Electron test skipped because Electron is not launchable here)
- Codex LSP diagnostics for `src/main`: 0 errors
- `node scripts/check-i18n.mjs` (exits 0; reports existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`)
- `git diff --check`

## Remote Checks
- `claude-review` failed before running review logic because the action could not fetch an OIDC token: `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable`; log suggests adding `id-token: write` to workflow permissions.

## Risks / Notes
- Persistence now returns to callers before disk rename completes; pending/in-flight writes are flushed on app quit.
- If an async disk write fails, behavior matches the prior non-throwing save path by logging the error and preserving in-memory state.

### DIFF
--- a/scripts/test-notes-async-saves.mjs
+++ b/scripts/test-notes-async-saves.mjs
@@ -1,0 +1,251 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import fsp from 'node:fs/promises';
+import Module, { createRequire } from 'node:module';
+import os from 'node:os';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { transform } from 'esbuild';
+
+const require = createRequire(import.meta.url);
+const repoRoot = process.cwd();
+const notesStorePath = path.join(repoRoot, 'src/main/notes-store.ts');
+
+function createMetrics() {
+  return {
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #500, opened 2026-07-03
perf(raycast): improve abortable promise hook lifecycle

JustYannicc:
## Summary
- Add real AbortController lifecycle handling to usePromise and useCachedPromise when callers pass options.abortable.
- Abort the current controller before starting a new run, abort active work on unmount, and clear abortable refs only when they still point at the matching run.
- Ignore late results/errors from older abortable runs so an aborted stale promise cannot overwrite the latest successful result.
- Add focused regression coverage for abort-on-revalidate, abort-on-unmount, and latest-result behavior across both hooks.

## Root Cause
`usePromise` accepted `abortable` but never created or published an `AbortController`. `useCachedPromise` created a controller for each run, but did not abort the previous controller, abort on unmount, clear refs, or guard against stale completions after a newer run started.

## Before / After Evidence
- Before: `node --test scripts/test-abortable-promise-hooks.mjs` failed 6/6 against the unmodified hooks. `usePromise` never exposed a controller; `useCachedPromise` left previous/unmounted controllers un-aborted and allowed stale data to overwrite latest data.
- After: `node --test scripts/test-abortable-promise-hooks.mjs` passes 6/6.

## Checks
- `node --test scripts/test-abortable-promise-hooks.mjs`
- `pnpm test`
- `pnpm run check:i18n` (exits 0; reports pre-existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`)
- `pnpm run build:main`
- `pnpm run build:renderer`
- Codex LSP diagnostics on both changed hook files: no errors
- `git diff --check`

## Risks
- Abortable hook callers may observe `abortable.current` returning to `null` after the active run settles or unmounts; this matches the controller ownership lifecycle and avoids retaining stale controllers.
- Non-abortable calls intentionally keep existing behavior, including legacy race semantics.

### DIFF
--- a/scripts/test-abortable-promise-hooks.mjs
+++ b/scripts/test-abortable-promise-hooks.mjs
@@ -1,0 +1,378 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+let activeHost = null;
+
+const fakeReact = {
+  useCallback: (...args) => activeHost.useCallback(...args),
+  useEffect: (...args) => activeHost.useEffect(...args),
+  useMemo: (...args) => activeHost.useMemo(...args),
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #501, opened 2026-07-03
perf(inline-args): coalesce settings writes

JustYannicc:
## Summary
- Add an opt-in debounced settings mirror for `sc-cmd-args:` writes while keeping `localStorage` updates synchronous.
- Route no-view inline argument typing through the debounced path so rapid keystrokes coalesce before IPC/settings persistence.
- Flush the pending command-argument mirror before no-view launch so the final argument payload is persisted before the command is queued.
- Add focused regression coverage for coalescing, launch flush, and immediate submit-style writes canceling pending debounced sync.

## Root Cause
`useLauncherInlineArguments` called `writeJsonObject(getCmdArgsKey(...), values)` on every no-view inline argument change. `writeJsonObject` immediately mirrored `sc-cmd-args:` keys through `saveExtensionCommandArguments`, so every keystroke could trigger a settings-store save and broadcast.

## Before / After Evidence
Baseline before production edit, using a Node/esbuild probe against the real bundled `extension-preferences.ts` with 11 rapid `sc-cmd-args:` updates:
- Before: 11 keystrokes produced 11 settings-mirror writes in 1.7 ms; `localStorage` immediately held `search text`.
- After: 11 keystrokes produced 0 immediate settings writes and 1 coalesced settings write after the 250 ms debounce window, measured at 301.8 ms total; `localStorage` still immediately held `search text`.

Focused regression coverage:
- Rapid inline-style writes coalesce to one `saveExtensionCommandArguments` call with the final value.
- `flushCommandArgumentSettingsSync(key)` writes the latest pending value immediately and avoids a later duplicate timer write.
- An immediate command-argument write cancels any pending debounced write for that key.

## Validation
- `node --test scripts/test-inline-args-write-coalesce.mjs`
- `node --test 'scripts/test-*.mjs'` (27 pass, 1 skipped live Electron test)
- `pnpm run build:renderer`
- `node scripts/check-i18n.mjs` (completed; existing Italian locale gaps remain: `settings.ai.whisper.vocabulary`, `settings.extensions.appSearchScope`)
- Codex LSP diagnostics for touched TypeScript files: no errors
- `git diff --cached --check`
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit` still fails on existing repo-wide renderer errors outside this change, including `App.tsx`, `CameraExtension.tsx`, `LiquidGlassSurface.tsx`, launcher browser profile typing, i18n runtime circular type, Raycast API typing, and quicklink icon typing.

## Risks
- The settings mirror for no-view inline argument typing is now delayed by up to 250 ms unless launch/submit flushes it first; `localStorage` and UI state remain immediate.
- The debounce is scoped to callers that opt in for command-argument keys. Existing preference writes, script argument writes, and normal command argument submit writes remain immediate by default.

### DIFF
--- a/scripts/test-inline-args-write-coalesce.mjs
+++ b/scripts/test-inline-args-write-coalesce.mjs
@@ -1,0 +1,146 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { build } from 'esbuild';
+
+const bundledPreferences = await build({
+  entryPoints: ['src/renderer/src/utils/extension-preferences.ts'],
+  bundle: true,
+  write: false,
+  format: 'esm',
+  platform: 'browser',
+  logLevel: 'silent',
+});
+
+const bundledPreferencesUrl =
+  'data:text/javascript;base64,' + Buffer.from(bundledPreferences.outputFiles[0].text).toString('base64');
+
+let moduleCounter = 0;
+
+function wait(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function copyPayload(value) {
+  return JSON.parse(JSON.stringify(value));
+}
+
+async function loadPreferencesWithMocks() {
+  const storage = new Map();
+  const commandArgumentWrites = [];
+
+  globalThis.localStorage = {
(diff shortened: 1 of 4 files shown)


JustYannicc:

PR #505, opened 2026-07-03
perf(script-command): avoid full script reads for headers

JustYannicc:
## Summary
- Replace full script reads during discovery with bounded prefix reads for the first Raycast metadata lines.
- Cache shebang interpreter metadata on discovered script commands and reuse it during execution.
- Add a source-backed benchmark plus regression tests for metadata parsing, shebang execution, no-shebang fallback, and large scripts.

## Metrics
Benchmark: `node scripts/bench-script-command-discovery.mjs` with 40 script command files, each with a 1.0 MiB body.

| Scenario | Before | After |
| --- | ---: | ---: |
| Discovery time | 98.8ms | 16.4ms |
| Discovery runner bytes read | 40.0 MiB | 1.3 MiB |
| Cached execution shebang lookup time | 8.7ms | 4.4ms |
| Cached execution shebang lookup runner bytes read | 1.0 MiB | 0 B |

## Validation
- `node scripts/bench-script-command-discovery.mjs`
- `pnpm test` (29 passed, 1 existing live Electron skip)
- `pnpm run build:main`
- `pnpm run check:i18n` (exits 0; reports existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`)
- Codex LSP diagnostics on `src/main`: 0 errors

## Risks
- Header discovery now reads up to the first 120 lines with a 256 KiB cap. That preserves the existing 120-line metadata window for normal Raycast headers, but a script with extraordinarily long header lines could need its metadata moved closer to the top.

### DIFF
--- a/scripts/bench-script-command-discovery.mjs
+++ b/scripts/bench-script-command-discovery.mjs
@@ -1,0 +1,102 @@
+#!/usr/bin/env node
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { loadScriptCommandRunner } from './lib/script-command-runner-harness.mjs';
+
+const fileCount = Number(process.env.SUPERCMD_BENCH_SCRIPT_COUNT || 40);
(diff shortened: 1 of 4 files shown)


JustYannicc:

PR #506, opened 2026-07-03
perf(root-search): optimize command scoring

JustYannicc:
## What changed
- Added `scripts/test-root-search-perf.mjs`, a focused synthetic root-search perf harness that compares the legacy two-pass command candidate output against the optimized path.
- Reused the root command match kind/score from `rankCommands` instead of rescanning the same command fields in `useLauncherCommandModel`.
- Skipped the legacy `filterCommands` list for normal non-empty root searches, where display already comes from assembled ranked root sections. Calculator fallback queries still compute the legacy list.

## Why
- Root search was doing duplicated O(commands * fields * query terms) work on each non-empty query.
- The legacy filtered list is not the displayed source for root search results, and command candidates only need the final root scoring pass that feeds `scoreRootSearchCandidate`.

## Compatibility impact
- Intended behavior is unchanged: the perf harness asserts optimized command matches have the same command ids, match kinds, and match scores as the legacy two-pass candidate path for representative queries.
- No user-visible strings or i18n keys changed.
- Returned legacy list fields are still populated for empty queries and calculator fallback queries; normal root-search display continues to use ranked root sections.

## How tested
- `node scripts/test-root-search-ranking.mjs`
- `node scripts/test-root-search-perf.mjs`
- `node --test 'scripts/test-*.mjs'`
- `node scripts/check-i18n.mjs`
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit` (fails on existing repo type errors outside this change, including `App.tsx`, `CameraExtension.tsx`, `LiquidGlassSurface.tsx`, browser profile typing in `useLauncherCommandModel.ts`, raycast runtime files, and quicklink icon typing).

Perf numbers from `node scripts/test-root-search-perf.mjs` over 6,000 synthetic commands, 8 queries, 6 iterations:
- Pre-edit captured baseline: legacy filter + two-pass command scoring median `3570.81ms`; optimized estimate before score reuse `2190.89ms`.
- Post-edit focused run: legacy comparator median `3872.38ms`; optimized command scoring path median `1753.83ms`; speedup `2.21x`.
- Post-edit full `node --test` run: legacy comparator median `3747.75ms`; optimized command scoring path median `1661.77ms`; speedup `2.26x`.

### DIFF
--- a/scripts/test-root-search-perf.mjs
+++ b/scripts/test-root-search-perf.mjs
@@ -1,0 +1,324 @@
+#!/usr/bin/env node
+
+import assert from 'assert/strict';
+import fs from 'fs';
+import path from 'path';
+import { createRequire } from 'module';
+import { performance } from 'perf_hooks';
+import vm from 'vm';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+const moduleCache = new Map();
+const ROOT = process.cwd();
+
+const reactStub = {
+  createElement: (type, props, ...children) => ({ type, props: props || {}, children }),
+  Fragment: 'Fragment',
+};
+
+function makeComponent(name) {
+  return function StubComponent(props) {
+    return { type: name, props: props || {}, children: [] };
+  };
+}
+
+const lucideStub = new Proxy({ __esModule: true }, {
+  get(target, prop) {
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #514, opened 2026-07-03
perf(background-refresh): stabilize refresh timers

JustYannicc:
## What changed
- Added a pure background refresh timer helper that builds stable timer keys from command identity, path, mode, and interval.
- Updated `useBackgroundRefresh` to reconcile keyed timers instead of clearing every interval on command-list or callback updates.
- Kept the extension and inline-script tick callbacks intact, with refs for the latest `fetchCommands` and menu-bar activity checks.
- Added regression coverage for unchanged, added, removed, changed, and unmount-cleared background timers.

## Why
Inline script refreshes can call `fetchCommands()`, which updates app command state. The previous hook cleanup cleared and recreated every background interval whenever that command list changed, even when background command identity/path/mode/interval did not change.

Baseline measurement from the new focused test shows the legacy unchanged-update behavior as `created=2 cleared=2`. The keyed reconcile path for the same unchanged update reports `created=0 retained=2 cleared=0`.

## Compatibility impact
- No-view and menu-bar extension refresh behavior is preserved: interval ticks still run the extension bundle, hydrate preferences, skip missing prefs/args, and dispatch `sc-launch-extension-bundle` with background launch options only for `no-view` or active `menu-bar` commands.
- Inline script behavior is preserved: interval ticks still read stored args, skip missing required args, run with `background: true`, and refresh commands when the result mode is `inline`.
- Removed commands clear their timers, and changed identity/path/mode/interval restarts only the changed timer.

## How tested
- `node --test scripts/test-background-refresh-timer-diff.mjs` passed.
  - Legacy unchanged update metric: `created=2 cleared=2`.
  - Keyed unchanged update metric: `created=0 retained=2 cleared=0`.
- `node --test 'scripts/test-*.mjs'` passed: 30 pass, 1 skipped live Electron test.
- LSP diagnostics for `src/renderer/src/hooks/useBackgroundRefresh.ts` and `src/renderer/src/hooks/backgroundRefreshTimers.ts`: no diagnostics.
- `./node_modules/.bin/tsc -p tsconfig.main.json && cp src/main/emoji-data.json dist/main/emoji-data.json` passed.
- `./node_modules/.bin/vite build` passed.
- `node scripts/check-i18n.mjs` completed; it still reports existing Italian locale gaps unrelated to this no-string change.
- `./node_modules/.bin/tsc -p tsconfig.renderer.json` still fails on existing unrelated renderer type errors outside this change.
- `node scripts/build-native.mjs` still fails before this code path because the native addon cannot resolve `node-addon-api`.

### DIFF
--- a/scripts/test-background-refresh-timer-diff.mjs
+++ b/scripts/test-background-refresh-timer-diff.mjs
@@ -1,0 +1,230 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const rootDir = path.resolve(__dirname, '..');
+const helperPath = path.join(rootDir, 'src/renderer/src/hooks/backgroundRefreshTimers.ts');
+
+function loadHelperModule() {
+  const source = fs.readFileSync(helperPath, 'utf8');
+  const compiled = ts.transpileModule(source, {
+    compilerOptions: {
+      module: ts.ModuleKind.CommonJS,
+      target: ts.ScriptTarget.ES2020,
+    },
+    fileName: helperPath,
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #518, opened 2026-07-03
fix(types): make renderer typecheck clean

JustYannicc:
## What changed

- Corrected renderer TypeScript mismatches across launcher command/profile types, inline refs, settings handlers, and Raycast runtime helper types.
- Added an ambient declaration for the existing `liquid-glass-react` renderer import so the renderer project can typecheck without a missing-module diagnostic.
- Kept fixes behavior-preserving: local type narrowing, signature alignment, typo correction for a timer ref, and compatibility type exposure for existing runtime behavior.

## Why

Renderer performance PRs were inheriting existing repo-wide TypeScript noise. The baseline renderer project check failed with exit 2 and 46 diagnostics across 20 files, so unrelated work could not rely on a clean renderer typecheck signal.

## Compatibility impact

No behavior removal intended. The changes align TypeScript declarations with existing runtime behavior and data shapes. User-facing strings were not added, renamed, or removed.

## How tested

- Before: `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false` exited 2 with 46 diagnostics.
- After: `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false` exited 0 with 0 diagnostics.
- LSP: `src/renderer/src` scan plus targeted checks for `raycast-api/index.tsx`, `hooks/useLauncherCommandModel.ts`, `settings/AITab.tsx`, and `utils/quicklink-icons.tsx` reported no errors.
- `pnpm test` passed: 24 pass, 1 skipped because live Electron is not launchable here.
- `pnpm run build:renderer` passed.
- `pnpm run check:i18n` exited 0; it still reports pre-existing Italian locale missing keys for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`.

PR #520, opened 2026-07-03
perf(browser-search): index hot query path

JustYannicc:
## What changed

- Added token-prefix, trigram, per-entry search, and bookmark nickname lookup data to the browser entry index.
- Routed bookmark/history query candidate selection through indexed entry ID intersections, then kept the existing scoring and ordering checks as the source of truth.
- Added a real synthetic browser search performance harness covering 24k history/bookmark rows, 300 tabs, nickname matches, full-scan parity, and generous query thresholds.

## Why

Browser search was still scanning every history and bookmark entry for typed root queries even though an entry index existed. Large browser datasets could make each keystroke spend more than 100 ms in candidate generation.

## Compatibility impact

Search result scoring, dedupe, completion, and ordering are preserved by verifying indexed candidates with the existing full scoring path. The code falls back to full scans when the query shape cannot be represented safely by the index. No user-facing strings or i18n keys changed.

## How tested

- Baseline before editing on synthetic 24,000 entries and 300 tabs: ranked avg 150.94-199.77 ms; ordered avg 173.56-224.59 ms.
- Final harness on synthetic 24,001 entries and 300 tabs: index build 300.50 ms; ranked full-scan avg 133.60-160.05 ms; ranked indexed avg 1.44-14.03 ms; ordered full-scan avg 134.62-154.54 ms; ordered indexed avg 1.39-12.52 ms.
- `node --test scripts/test-browser-search-lag-fix.mjs scripts/test-browser-search-performance.mjs`
- `node --test 'scripts/test-*.mjs'`
- `node scripts/check-i18n.mjs`
- Codex LSP diagnostics for `src/renderer/src/hooks/useBrowserSearch.ts`: no errors.

## Stack validation

This issue was found and validated from `codex/perf-integration-stack`, which contains the current performance fix stack, including the renderer typecheck cleanup from #518. On that integrated branch, `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false`, `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit --pretty false`, and `node --test 'scripts/test-*.mjs'` pass.

### DIFF
--- a/scripts/test-browser-search-lag-fix.mjs
+++ b/scripts/test-browser-search-lag-fix.mjs
@@ -57,8 +57,12 @@ test('Browser search lag fix', async (t) => {
     assertIncludes(files.hook, 'refreshEntriesIfStale');
     assertIncludes(files.hook, 'historyByTimeEntryIds');
     assertIncludes(files.hook, 'bookmarksByBrowserOrderEntryIds');
+    assertIncludes(files.hook, 'tokenPrefixEntryIds');
+    assertIncludes(files.hook, 'tokenTrigramEntryIds');
+    assertIncludes(files.hook, 'getIndexedEntryCandidateIds');
     assertIncludes(files.hook, 'BROWSER_ENTRY_INDEX_MAX_TOKEN_LENGTH = 128');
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #524, opened 2026-07-03
perf(oauth): cap callback queue

JustYannicc:
## What changed

- Store OAuth callback queue entries as parsed callback records with received timestamps.
- Ignore malformed and non-OAuth callback URLs before they can enter the queue.
- Prune callbacks older than the existing OAuth callback timeout and cap retained unmatched callbacks at 256 entries.
- Add focused node:test coverage for 1000 unmatched callbacks, stale callback pruning, matching-state resolution, and malformed/non-OAuth callback handling.

## Why

Long-running app sessions could accumulate stale OAuth callback URLs because the bridge pushed every callback string into an unbounded queue and unmatched entries were retained indefinitely. The hot path is the renderer OAuth callback listener, where stale callbacks should remain available only for the existing callback wait window and should never grow without bound.

## Compatibility impact

Raycast OAuth compatibility is preserved: valid SuperCmd OAuth callback URLs still parse the same fields, queued matching states still resolve, and live waiters still resolve immediately. Non-OAuth or unparsable URLs are ignored safely instead of being retained. No user-facing strings or i18n files changed.

## How tested

- Baseline before editing: on the stack-based worker branch, code inspection showed `oauthCallbackQueue: string[]` with unconditional `push` and no cap or age pruning; the new regression uses 1000 unmatched callbacks to cover that previously unbounded retention path.
- After fix: `node --test scripts/test-oauth-callback-queue.mjs` passed on the scoped PR branch, 4/4 tests.
- Targeted TypeScript check: `./node_modules/.bin/tsc --noEmit --target ES2020 --lib ES2020,DOM,DOM.Iterable --module ESNext --moduleResolution bundler --jsx react-jsx --strict --esModuleInterop --skipLibCheck --forceConsistentCasingInFileNames --resolveJsonModule --isolatedModules src/renderer/src/raycast-api/oauth/oauth-bridge.ts` passed.
- Renderer typecheck on scoped `origin/main` branch: `./node_modules/.bin/tsc -p tsconfig.renderer.json` still fails on existing unrelated main-branch errors outside this PR, including `src/renderer/src/App.tsx`, `src/renderer/src/CameraExtension.tsx`, `src/renderer/src/components/LiquidGlassSurface.tsx`, and launcher/browser type mismatches. The stack-based implementation branch passed this renderer typecheck before the scoped cherry-pick.

## Stack validation

The issue was analyzed and implemented from `codex/perf-integration-stack`, then moved to scoped branch `codex/perf-oauth-callback-queue-cap-pr` from `origin/main` by cherry-picking only the OAuth fix commit. The upstream diff is limited to `src/renderer/src/raycast-api/oauth/oauth-bridge.ts` and `scripts/test-oauth-callback-queue.mjs`; no coordination or playbook markdown files are included.

### DIFF
--- a/scripts/test-oauth-callback-queue.mjs
+++ b/scripts/test-oauth-callback-queue.mjs
@@ -1,0 +1,141 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { build } from 'esbuild';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const bridgePath = path.join(root, 'src/renderer/src/raycast-api/oauth/oauth-bridge.ts');
+let importNonce = 0;
+
+async function importOAuthBridge() {
+  const result = await build({
+    entryPoints: [bridgePath],
+    bundle: true,
+    write: false,
+    format: 'esm',
+    platform: 'node',
+    target: 'node20',
+    logLevel: 'silent',
+  });
+  const code = result.outputFiles[0].text;
+  const dataUrl = `data:text/javascript;base64,${Buffer.from(code).toString('base64')}#oauth-${importNonce++}`;
+  return import(dataUrl);
+}
+
+async function loadBridge(t) {
+  const previousWindow = globalThis.window;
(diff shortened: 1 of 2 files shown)


JustYannicc:

PR #529, opened 2026-07-03
perf(camera): avoid duplicate capture encoding

JustYannicc:
## What changed
- Moved camera capture encoding into a small camera-specific helper module.
- Replaced the capture preview data URL with an object URL created from the same PNG blob used for disk save.
- Revoke the preview object URL when the preview clears and when the camera component unmounts.
- Added a focused regression test for the legacy duplicate encode baseline and the new single-encode path.

## Why
The previous capture flow encoded the same canvas frame twice: once with `toDataURL` for preview and once with `toBlob` for saving. Data URLs also inflate memory. The new flow performs one `toBlob` encode, uses that blob for both preview and save, and keeps the saved file behavior unchanged.

## Compatibility impact
- Camera preview UX is preserved: the latest capture thumbnail is still shown and fades/clears on the same timers.
- Saved PNG file behavior, save path, clipboard write, and last-capture open behavior are unchanged.
- Preview memory lifetime is improved because object URLs are explicitly revoked on clear and unmount.

## How tested
- `node --test scripts/test-camera-capture-single-encode.mjs`
  - Legacy baseline mock confirms one capture would call `toDataURL` once and `toBlob` once.
  - Production helper path confirms `toBlob` is called once and `toDataURL` is not called.
  - Confirms object URL preview is shown and revoked on clear/dispose.
- `./node_modules/.bin/tsc --noEmit --jsx react-jsx --target ES2020 --module ESNext --moduleResolution bundler --strict --esModuleInterop --skipLibCheck --lib ES2020,DOM,DOM.Iterable src/renderer/types/electron.d.ts src/renderer/src/camera-capture.ts src/renderer/src/CameraExtension.tsx`
- `git diff --check origin/main...HEAD -- src/renderer/src/CameraExtension.tsx src/renderer/src/camera-capture.ts scripts/test-camera-capture-single-encode.mjs`

Full renderer typecheck was also attempted with `./node_modules/.bin/tsc -p tsconfig.renderer.json --pretty false`, but `origin/main` currently fails on unrelated renderer errors in files such as `src/renderer/src/App.tsx`, `src/renderer/src/components/LiquidGlassSurface.tsx`, launcher hooks, Raycast runtime files, settings, and quicklink icon utilities. No camera-file errors were reported by the scoped camera typecheck above.

No physical camera timing numbers were produced because this worker environment does not provide an Electron camera capture session/hardware harness. The regression test provides fresh encoder call-count validation for the performance issue.

## Stack validation
- Branch `codex/perf-camera-single-encode-pr` was created from `origin/main`.
- `git rev-list --left-right --count origin/main...HEAD` reports `0 1`.
- `git log origin/main..HEAD` contains only `70b0e5c perf(camera): avoid duplicate capture encoding`.
- Diff is scoped to `src/renderer/src/CameraExtension.tsx`, `src/renderer/src/camera-capture.ts`, and `scripts/test-camera-capture-single-encode.mjs`.

### DIFF
--- a/scripts/test-camera-capture-single-encode.mjs
+++ b/scripts/test-camera-capture-single-encode.mjs
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { importTs } from './lib/ts-import.mjs';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const cameraExtensionPath = path.join(root, 'src/renderer/src/CameraExtension.tsx');
+const cameraCapturePath = path.join(root, 'src/renderer/src/camera-capture.ts');
+
+const {
+  createCapturePreview,
+  createCapturePreviewUrlManager,
+  encodeCanvasAsPngBlob,
+} = await importTs(cameraCapturePath);
+
+function makeMockCanvas(blob = new Blob(['fake-png'], { type: 'image/png' })) {
+  const calls = {
+    toBlob: [],
+    toDataURL: [],
+  };
+
+  return {
+    calls,
+    toBlob(callback, type) {
+      calls.toBlob.push(type);
+      callback(blob);
+    },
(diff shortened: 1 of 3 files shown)


JustYannicc:

PR #522, opened 2026-07-03
perf(exec): clear command timeout handles

JustYannicc:
## What changed

- Moved the `exec-command` runner into `src/main/exec-command.ts` so its process/timer lifecycle can be exercised without booting Electron.
- Added a single guarded finish path that clears the 300000ms timeout for close, error, and timeout completion paths and ignores late duplicate events.
- Added regression coverage for close-before-timeout, error-before-timeout, timeout kill behavior, and close/error/timeout races.

## Why

Every `exec-command` invocation created a 5-minute timeout, but successful close and spawn-error paths never cleared it. A baseline fake child-process harness reproduced the leak with `activeTimeouts=1` after both close-before-timeout and error-before-timeout.

## Compatibility impact

The IPC result shape and behavior are preserved: close still returns collected stdout/stderr with the exit code, error still returns the error message with exit code 1, and timeout still kills the process and returns exit code 124 with the previous timeout stderr behavior.

## How tested

- Baseline before edit: fake child-process harness showed `activeTimeouts=1` after close-before-timeout and error-before-timeout.
- After fix: `node --test scripts/test-exec-command-timeout-cleanup.mjs` confirms close/error paths clear the timer and timeout/race paths settle once.
- `node --test scripts/test-exec-command-timeout-cleanup.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json`
- `node --test scripts/test-exec-command-timeout-cleanup.mjs scripts/test-browser-search-lag-fix.mjs scripts/test-renderer-crash-recovery.mjs scripts/test-renderer-error-boundary.mjs scripts/test-root-search-ranking.mjs`
- Codex LSP diagnostics on `src/main/exec-command.ts` and `src/main/main.ts`: no errors.

## Stack validation

This issue was found and validated from `codex/perf-integration-stack`, which contains the current performance fix stack. The PR branch is intentionally scoped against `origin/main`; the same patch was also applied cleanly to the integration stack and validated there with `./node_modules/.bin/tsc -p tsconfig.main.json` plus focused stack tests.

PR #512, opened 2026-07-03
perf(background): dedupe no-view extension runs

JustYannicc:
## Summary
- Add a shared background no-view run queue helper that dedupes background launches by extension/command identity while preserving repeated user-initiated launches.
- Wire App queueing through the helper and keep hidden runner cleanup on success/error via onClose plus queue clearing if the hidden runner container unmounts.
- Add a focused regression simulation for overlapping ticks, mixed legacy/hydrated identity fields, explicit user launches, and cleanup/re-enqueue paths.

## Root Cause
Background interval ticks dispatch hidden no-view extension bundles without checking whether the same extension command is already in flight, so long-running commands can accumulate duplicate hidden ExtensionView runners.

## Before / After Evidence
- Baseline simulation before behavior edit: `[baseline] background no-view duplicate ticks: ticks=3 intervalMs=25 elapsedMs=50 queued=3`.
- After helper wiring: `[after] background no-view duplicate ticks: ticks=3 intervalMs=25 elapsedMs=50 queued=1 skipped=2`.

## Validation
- `node --test scripts/test-background-no-view-dedupe.mjs` passes 6 tests.
- `node --test scripts/test-*.mjs` passes 30 tests, skips 1 live Electron test.
- `node scripts/check-i18n.mjs` exits 0; it still reports the existing non-strict `it` locale missing keys for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`.
- `./node_modules/.bin/tsc -p tsconfig.main.json` passes.
- `./node_modules/.bin/vite build` passes.
- Codex LSP diagnostics are clean on `src/renderer/src/utils/background-no-view-runs.ts`, `src/renderer/src/components/HiddenExtensionRunners.tsx`, and `src/renderer/src/hooks/useMenuBarExtensions.ts`.

## Known Limitations / Risks
- `./node_modules/.bin/tsc -p tsconfig.renderer.json` still fails on pre-existing project-wide renderer type errors outside this patch, including `src/renderer/src/App.tsx:759` and `src/renderer/src/App.tsx:1081`.
- `node scripts/build-native.mjs` is blocked in this local environment because `npx` is disabled and the fallback `pnpm dlx node-gyp` cannot resolve `node-addon-api` for `src/native/native-helpers-addon`.
- `pnpm test` is blocked by the local pnpm install/build-script approval hook, so validation used the underlying Node test command directly.

PR #510, opened 2026-07-03
perf(cache): maintain Raycast cache sizes incrementally

JustYannicc:
## Summary
- Store Raycast Cache metadata as versioned `lruOrder`, `sizeByKey`, and `totalSize` data.
- Update sizes incrementally on set/get/remove/clear/eviction so eviction no longer rescans every cached localStorage item in a loop.
- Recover/migrate legacy `{ lruOrder }` metadata, missing metadata, corrupt JSON metadata, and malformed LRU metadata by rebuilding sizes from cache item keys.
- Add a focused Node harness that exercises cache behavior, LRU eviction, metadata migration/recovery, and prints localStorage read/time metrics.

## Before/After Evidence
Focused benchmark: 120 entries, 1 KiB values, eviction from 120 entries down to a 60-entry capacity.

| Scenario | Before getItem reads | After getItem reads | Before duration | After duration |
| --- | ---: | ---: | ---: | ---: |
| set-fill | 7,140 | 0 | 2.741 ms | 1.269 ms |
| get-existing | 120 | 120 | 0.371 ms | 2.982 ms |
| set-with-eviction | 5,549 | 0 | 1.238 ms | 0.102 ms |

Durations are local-run timings, so the read counts are the primary signal.

## Validation
- `node --test scripts/test-raycast-cache-size-index.mjs` passes; latest after metrics: set-fill `getItem=0`, get-existing `getItem=120`, set-with-eviction `getItem=0`.
- `pnpm run test` passes: 30 passed, 1 skipped.
- `pnpm run check:i18n` exits 0; existing Italian missing keys reported: `settings.ai.whisper.vocabulary`, `settings.extensions.appSearchScope`.
- `pnpm run build:main` passes.
- `pnpm run build:renderer` passes.
- GitHub `claude-review` check fails before review because the workflow cannot fetch an OIDC token: missing `ACTIONS_ID_TOKEN_REQUEST_URL` / likely needs `id-token: write` workflow permission.
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit` still fails on existing project-wide renderer type errors unrelated to this cache change.
- Codex LSP diagnostics on `src/renderer/src/raycast-api/index.tsx` still report existing `Toast.Style` merge/type errors and `addHexAlpha` typing outside the cache block.

## Risks / Notes
- If cache metadata is absent or corrupt, the first constructor recovery may scan matching localStorage cache item keys once, then saves repaired metadata.
[...]

PR #509, opened 2026-07-03
perf(extension): reduce sync IPC in extension facades

JustYannicc:
## What changed
- Added a focused extension sync-facade measurement/regression script for representative real and virtual filesystem path operations.
- Added a lazy Node-backed sync adapter in `ExtensionView.tsx` for real filesystem and child_process sync fallbacks when Node require is available.
- Kept virtual/localStorage filesystem reads ahead of real filesystem access and preserved the existing `window.electron` sync IPC fallback for non-Node contexts.
- Routed `fs`, `fs/promises`, and `child_process` module resolution through SuperCmd facades so extensions keep the compatibility layer instead of raw Node bypassing it.

## Why
Extension filesystem and shell-like facades were using preload `sendSync` calls for real path lookups and sync child process calls. Those calls block the renderer and main process even though the launcher renderer already has Node APIs available.

## Metrics
Focused harness: `node scripts/test-extension-sync-facade.mjs`.

| Scenario | Real path sync IPC calls | Breakdown | Sync IPC time |
| --- | ---: | --- | ---: |
| Before production edit, fallback path | 9 | execCommandSync=1, fileExistsSync=2, readFileSync=1, statSync=5 | 0.318ms |
| After, fallback path still available | 9 | execCommandSync=1, fileExistsSync=2, readFileSync=1, statSync=5 | 0.374ms |
| After, Node-backed path | 0 | execCommandSync=0, fileExistsSync=0, readFileSync=0, statSync=0 | 0.000ms |

Virtual/localStorage precedence stayed at 0 sync IPC calls before and after.

## Validation
- Baseline before implementation: `node scripts/test-extension-sync-facade.mjs --mode=fallback`
- `node scripts/test-extension-sync-facade.mjs`
- `node --test scripts/test-*.mjs` (passes; live Electron crash-recovery test skipped here)
- `node scripts/check-i18n.mjs` (exits 0; existing Italian missing-key warnings remain for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`)
- Codex LSP diagnostics on `src/renderer/src/ExtensionView.tsx`: no errors
- `node_modules/.bin/tsc -p tsconfig.main.json`
- `node_modules/.bin/vite build`
- `git diff --check`

## Notes and risks
[...]

PR #530, opened 2026-07-03
perf(runtime): reduce extension and menubar churn

JustYannicc:
## What changed
- Consolidates runtime, background refresh, menu-bar, cache, OAuth, exec, and hook lifecycle performance fixes into one upstream PR.
- Adds bounded/incremental caches for `withCache`, Raycast `Cache`, extension wrapper compilation, extension bundle preparation, and menu-bar visible payload updates.
- Reduces long-running menu-bar churn by avoiding storage self-remounts, deduping background no-view runs, and retaining unchanged background refresh timers.
- Reduces extension runtime overhead by preferring in-process Node facades for sync `fs`/`child_process` paths and by clearing abortable promise and exec timeout handles.
- Caps unmatched OAuth callback retention and coalesces inline command argument settings writes.
- Adds focused regression and measurement scripts for the touched runtime paths.

## Why
Long-lived menu-bar sessions can accumulate work from repeated refresh timer recreation, unchanged tray payload IPC, storage-triggered remounts, background no-view overlap, cache growth, sync IPC, and callback queues. This PR combines the related fixes so the upstream review can evaluate the full lag-reduction path together.

## Compatibility impact
- Raycast-compatible behavior is preserved: external and sibling storage changes still refresh menu-bar commands, while a menu-bar command no longer remounts itself for its own local storage write.
- User-initiated no-view launches are still allowed to overlap; only background ticks for the same extension command are deduped while a run is in flight.
- Extension wrapper and bundle caches invalidate on code or file-stat changes, and install/uninstall/build paths clear runner caches.
- Legacy Raycast cache metadata is migrated or repaired, and per-key `has`/`remove` still reconcile stale localStorage entries.
- OAuth callback semantics remain state-based, with stale/unmatched callbacks pruned to avoid unbounded queue growth.

## How tested
- `node --test scripts/test-with-cache-expiry.mjs scripts/test-menubar-payload-cache.mjs scripts/test-extension-wrapper-cache.mjs scripts/test-extension-bundle-cache.mjs scripts/test-menubar-storage-remounts.mjs scripts/test-abortable-promise-hooks.mjs scripts/test-inline-args-write-coalesce.mjs scripts/test-extension-sync-facade.mjs scripts/test-raycast-cache-size-index.mjs scripts/test-background-no-view-dedupe.mjs scripts/test-background-refresh-timer-diff.mjs scripts/test-exec-command-timeout-cleanup.mjs scripts/test-oauth-callback-queue.mjs`
- `node scripts/measure-extension-wrapper-cache.mjs`
- `node scripts/measure-extension-bundle-cache.mjs`
- `pnpm exec tsc -p tsconfig.main.json --noEmit --pretty false`
- `pnpm test`
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit --pretty false` fails on the standalone branch because of existing repo-wide renderer typing debt unrelated to this consolidation.

## Performance evidence
- Menu-bar equivalent visible payload registrations drop from 3 sends to 1 send.
- Background refresh unchanged updates drop from legacy `created=2 cleared=2` to keyed `created=0 retained=2 cleared=0`.
- Background no-view duplicate ticks drop from `queued=3` to `queued=1 skipped=2`.
- Extension wrapper compilation drops from 1000 wrapper creations to 1 cached wrapper creation in the focused harness.
- Raycast cache benchmark avoids full scans on set/eviction (`set-fill getItem=0`, `set-with-eviction getItem=0`).
- Extension sync facade Node path reports 0 sync IPC calls for real path ops in the focused harness.

## Stack validation
Applied this consolidation over `codex/consolidate-validation-checks` (`fix(validation): restore project checks`) and validated the combined stack:
- `pnpm test`

## Replaces
- #482
- #486
- #490
- #498
- #499
- #500
- #501
- #509
- #510
- #512
- #514
- #522
- #524

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,40 @@
+name: Test
+
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'scripts/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/test.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/**'
+      - 'scripts/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/test.yml'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run node test suite
+        run: npm test
--- a/scripts/measure-extension-bundle-cache.mjs
+++ b/scripts/measure-extension-bundle-cache.mjs
@@ -1,0 +1,257 @@
+#!/usr/bin/env node
+
+import { fileURLToPath, pathToFileURL } from 'node:url';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { createRequire } from 'node:module';
+import { build } from 'esbuild';
+import { performance } from 'node:perf_hooks';
+
+const require = createRequire(import.meta.url);
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const iterations = Number.parseInt(process.env.SUPERCMD_EXTENSION_BUNDLE_MEASURE_ITERATIONS || '20', 10);
+
+export function createFixture() {
+  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'sc-extension-bundle-cache-'));
+  const userDataDir = path.join(tmpDir, 'userData');
+  const extensionsDir = path.join(userDataDir, 'extensions');
+  const extDir = path.join(extensionsDir, 'bundle-cache-fixture');
+  const buildDir = path.join(extDir, '.sc-build');
+
+  fs.mkdirSync(buildDir, { recursive: true });
+  fs.writeFileSync(
+    path.join(extDir, 'package.json'),
+    JSON.stringify(
+      {
+        name: 'bundle-cache-fixture',
+        title: 'Bundle Cache Fixture',
+        description: 'Fixture for extension bundle cache measurement',
+        owner: 'codex',
+        commands: [
+          {
+            name: 'background',
+            title: 'Background',
+            description: 'No-view command',
+            mode: 'no-view',
+            preferences: [
+              {
+                name: 'freshCommandPref',
+                title: 'Fresh Command Pref',
+                type: 'textfield',
+                default: 'default-command',
+              },
+            ],
+          },
+          {
+            name: 'menu',
+            title: 'Menu',
+            description: 'Menu-bar command',
+            mode: 'menu-bar',
+          },
+        ],
+        preferences: [
+          {
+            name: 'freshExtensionPref',
+            title: 'Fresh Extension Pref',
+            type: 'textfield',
+            default: 'default-extension',
+          },
+        ],
+      },
+      null,
+      2
+    )
+  );
+  fs.writeFileSync(path.join(buildDir, 'background.js'), 'module.exports = { name: "background" };\n');
+  fs.writeFileSync(path.join(buildDir, 'menu.js'), 'module.exports = { name: "menu" };\n');
+
+  return { tmpDir, userDataDir, extDir };
+}
+
+export async function bundleExtensionRunner(userDataDir) {
+  const outFile = path.join(
+    os.tmpdir(),
+    `sc-extension-runner-measure-${process.pid}-${Date.now()}.cjs`
+  );
+
+  const stubModule = (filter, source) => ({
+    name: `stub-${filter}`,
+    setup(pluginBuild) {
+      pluginBuild.onResolve({ filter }, () => ({
+        path: filter.source,
+        namespace: `stub-${filter.source}`,
+      }));
+      pluginBuild.onLoad({ filter: /.*/, namespace: `stub-${filter.source}` }, () => ({
+        contents: source,
+        loader: 'js',
+      }));
+    },
+  });
+
+  await build({
+    entryPoints: [path.join(root, 'src/main/extension-runner.ts')],
+    outfile: outFile,
+    bundle: true,
+    platform: 'node',
+    format: 'cjs',
+    target: 'node20',
+    logLevel: 'silent',
+    plugins: [
+      stubModule(
+        /^electron$/,
+        `
+          module.exports = {
+            app: {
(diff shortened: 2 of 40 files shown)


shobhit99:
This PR consolidates 13 upstream perf PRs: bounded/incremental caches (Raycast `Cache` size index, extension manifest/bundle/wrapper caches, `withCache` eviction), key-stable background-refresh timers, menu-bar payload dedup + self-remount suppression, background no-view dedup, abortable-promise/exec-timeout cleanup, an OAuth callback queue bound, in-process Node `fs`/`child_process` sync facades, and debounced command-argument settings sync, plus ~3k lines of focused `node --test` scripts.

Reviewed the full diff across all 20 source files. Overall this is careful, well-tested work that is behavior-preserving in most areas. One change stands out as a genuine behavior shift worth verifying before merge.

## Strengths
- **exec-command extraction** (`src/main/exec-command.ts`) is a faithful move of the inline `run-command` IPC handler, and it fixes a latent issue by clearing the 5-min timeout on completion (base left it armed, so it would `proc.kill()` a dead process later).
- **extension-runner stat-signature caches** invalidate correctly: build/single-build paths call `invalidateExtensionStaticCacheForPath`, and install/uninstall/build in `extension-registry.ts` call the full `invalidateExtensionRunnerCaches()`. Verified no caller mutates the now-shared cached `pkg` object, so the shared-reference manifest cache is safe.
- **usePromise / useCachedPromise** now abort superseded runs and guard stale state writes via `isCurrentRun()` — this is *more* correct than base (which orphaned in-flight runs and could let an old resolution overwrite new data), not merely faster. Abort-on-unmount is guarded so the resulting AbortError doesn't surface.
- **OAuth callback queue** (`oauth-bridge.ts`): parse-once + "handled by waiter" dedup + prune/cap fixes a real base bug where consumed URLs were left in the queue and the queue could grow unbounded.
- **Raycast `Cache`**: O(1) size tracking via incremental `sizeByKey`/`totalSize`, with a proper v1→v2 metadata migration and reconciliation when localStorage is mutated externally. Nicely done.
- No i18n impact — no user-facing strings were added/renamed/removed (only `console.*` logs).

## Concerns / potential bugs
1. **(Primary — please verify before merge) `fs` / `child_process` now resolve to the SuperCmd stub facade even in Node-enabled extension hosts.** In `ExtensionView.tsx`, `loadExtensionExport`'s `fakeRequire` gained a `shouldUseSuperCmdBuiltinFacade(name)` early-return that hands back `fsStub` / `childProcessStub` for `fs`, `fs/promises`, and `child_process` **before** `tryRealNodeRequire(name)`. Base returned the *real* Node modules first, and real Node require is available here (confirmed via `_realNodeRequire` / `window.__scNodeRequire`). `fsStub`/`childProcessStub` are plain object literals with **no fallthrough** to the real module, so anything they don't implement is absent or a no-op stub: `createReadStream`/`createWriteStream` are stubs (not real disk streams), and `watch`, `cpSync`, `mkdtempSync`, `appendFileSync`, `readlinkSync`/`symlinkSync`, `opendir`, `chmodSync`, the `Dirent`/`Stats` classes, and real streaming async `child_process.spawn`/`exec` are missing. Extensions relying on real streaming fs or real child processes in a Node-enabled window could regress. The common sync `read`/`stat`/`readdir`/`exists` paths *do* delegate to real fs (via `getRealNodeFsModule()`), so this is likely intentional (asset-path layering + sync-IPC reduction) and fine for the majority — but I'd (a) confirm this matches the intended node-integration model, and (b) consider making the facades fall through to the real module for members they don't implement (e.g. a `Proxy` backed by the real module, or spreading it).

2. **(Low) Leftover per-load debug log.** `getCompiledExtensionWrapper`'s call site logs `console.debug('[loadExtensionExport] Wrapper cache hit/miss …')` on every extension load. Drop it or gate it behind a debug flag.

3. **(Low) Menu-bar remount throttle reorder** (`useMenuBarExtensions.ts`, `remountMenuBarExtensionsForExtension`). The 200ms throttle timestamp is now written *inside* the `setState` updater and only `if (changed)`. Two `sc-extension-storage-changed` events firing synchronously in the same tick (before React flushes the updater) can both pass the `now - lastTs < 200` guard and both remount, where base recorded the timestamp synchronously and would throttle the second. Net effect is an occasional extra remount (mildly ironic for a churn-reduction PR) — not incorrect. Also, mutating a ref inside the updater is a side-effect-in-reducer pattern (double-invoked under StrictMode, though idempotent here).

4. **(Low, theoretical) Wrapper cache key collisions.** `extension-wrapper-cache.ts` keys on `extensionIdentity + code.length + 32-bit FNV-1a(code)`. A same-identity, same-length, same-hash collision would return a stale wrapper. Astronomically unlikely per-extension; noting for completeness.

5. **(Low) `Cache.has()` gained side effects.** On the size-mismatch reconciliation path it calls `updateLRU` + `saveToStorage`, so `has()` can bump LRU recency (Raycast's `has` is conceptually a pure existence check). Only triggers on external localStorage mutation, so minor.

## Suggestions
- Confirm CI actually executes the new `scripts/test-*.mjs` (`node --test`) suite — otherwise the strongest safety net for this consolidation isn't running in the pipeline.
- This is a large, cross-subsystem PR (41 files, +5k/-377) folding 13 PRs together. ~3k lines are tests (good), but a regression in the source will be hard to bisect/revert granularly; concern #1 in particular would be easier to reason about as its own commit.

🟡 Approve with minor comments — the work is careful and well-tested; concern #1 (fs/child_process facade preference) is the one behavior change I'd want confirmed against real extensions before merging.

JustYannicc:
Addressed the actionable review feedback in `11b7b72`:

- `fs` / `child_process` facade priority/fallthrough: kept the SuperCmd overlay for virtual/localStorage-aware behavior, but `fakeRequire` now returns a proxy-backed facade in Node-enabled hosts so missing members fall through to real Node modules. Placeholder fs methods that previously blocked real Node (`openSync`, `readSync`, `writeSync`, `closeSync`, `rmdirSync`, stream/watch helpers, etc.) now delegate to real `fs` when available, and `child_process.fork` delegates to real Node when available.
- Wrapper-cache debug logging: gated the per-load `[loadExtensionExport] Wrapper cache ...` log behind `globalThis.__SUPERCMD_DEBUG_EXTENSION_WRAPPER_CACHE` so normal extension loads stay quiet.
- Menu-bar remount throttle: timestamping is synchronous before queued React state updates, but only after confirming there is a real remount target. Added a regression test for a self-originated no-op write followed immediately by an external refresh.
- Test-suite execution: added `.github/workflows/test.yml` so PRs/pushes touching root source/scripts/package metadata run `npm ci` + `npm test`.

Verification:
- `node --test scripts/test-background-no-view-dedupe.mjs scripts/test-background-refresh-timer-diff.mjs scripts/test-extension-sync-facade.mjs scripts/test-menubar-storage-remounts.mjs scripts/test-oauth-callback-queue.mjs` ✅ 24 passed
- `git diff --check` ✅

shobhit99 (approved):

PR #532, opened 2026-07-03
fix(validation): restore project checks

JustYannicc:
## What changed

Consolidates the validation/build fixes from these source PRs into one upstream PR:

- Replaces #494 `ci: add project regression checks`
- Replaces #496 `test: add bundled TypeScript import helper`
- Replaces #518 `fix(types): make renderer typecheck clean`
- Replaces #519 `ci(claude): fix review action on fork PRs`
- Replaces #521 `fix(i18n): restore Italian parity checks`
- Replaces #528 `fix(pnpm): approve native build scripts`

The combined change adds a root Project Checks workflow, upgrades the test TypeScript import helper, makes the renderer project typecheck clean, skips Claude review on fork PRs that cannot receive OIDC/secrets, restores strict i18n parity with the missing Italian keys, and approves the native pnpm build scripts needed for local validation.

## Why

The individual validation PRs were fixing related gaps in the project safety net: missing root project checks, unreliable TypeScript import coverage in tests, renderer typecheck noise, fork PR review-check failures, non-strict locale drift, and pnpm 11 native-build approvals. Keeping them together gives upstream one coherent PR that restores the expected validation path for follow-on work.

## Compatibility impact

Most changes are CI, test, type, or install metadata only. The runtime-facing changes are behavior-preserving type alignments and added Italian translation strings. Same-repository PRs still run Claude review; fork PRs skip that job because the required credentials are not available to `pull_request` runs. npm installs remain unchanged, while pnpm installs may now run the explicitly approved native build scripts.

## How tested

- `pnpm install` passed and ran the approved native postinstall/rebuild path.
- Codex LSP diagnostics: renderer source scan reported 0 errors; focused `raycast-api` and `hooks` scans also reported 0 errors.
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit --pretty false` exited 0.
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false` exited 0 with no diagnostics.
- `pnpm exec tsc -p tsconfig.main.json --noEmit --pretty false` exited 0.
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit --pretty false` exited 0 with no diagnostics.
- `pnpm run check:i18n` passed for de, es, fr, it, ja, ko, ru, zh-Hans, and zh-Hant.
- `pnpm test` passed: 28 tests, 0 failures.
- `pnpm run build:main` passed.
[...]

PR #526, opened 2026-07-03
perf(icons): coalesce file icon IPC

JustYannicc:
## What changed

- Added path-keyed in-flight coalescing around Raycast file icon IPC requests so concurrent identical file icons share one `getFileIconDataUrl` call.
- Replaced the unbounded renderer `fileIconCache` with a 1024-entry LRU cache for successful data URLs and resolved null fallbacks.
- Kept rejected IPC failures retryable by avoiding permanent negative cache writes on rejection.
- Added a self-contained `node:test` harness for concurrent duplicate icons, rejection recovery, LRU eviction, and non-file image icon behavior.

## Why

The file icon runtime lived on a hot render path for Raycast list/grid items. Before this change, duplicate mounted file icons could all observe a cache miss before the first IPC call resolved, so identical icons fanned out into repeated `getFileIconDataUrl` IPC calls. The cache was also a module-level `Map` with no cap, so browsing or rendering many unique paths retained every icon result for the life of the renderer.

Baseline mock-render harness for the pre-fix request/cache logic reported `sameMountedIcons=100`, `sameIpcCalls=100`, `uniqueMountedIcons=5000`, `uniqueIpcCalls=5000`, and `uniqueCacheEntries=5000`.

## Compatibility impact

Raycast icon rendering semantics are preserved. File icons still render the same data URL image when available, and still fall back to the existing document/folder glyph while loading or after unavailable icons. Cached file icons may remain stale, matching the existing tolerance. Non-file icon paths continue through the existing remote/image/Phosphor handling. No user-facing strings or locale files changed.

## How tested

- Baseline before fix: `node -e <mock old FileIcon request/cache harness>` -> `{"sameMountedIcons":100,"sameIpcCalls":100,"uniqueMountedIcons":5000,"uniqueIpcCalls":5000,"uniqueCacheEntries":5000}`.
- After fix on scoped branch: `node --test scripts/test-icon-runtime-file-icon-cache.mjs` -> pass; asserts 100 concurrent identical mounts call IPC once, rejected IPC retries later, 5,000 unique paths stay capped at 1,024 cache entries with LRU eviction, and non-file image icons do not call file icon IPC.
- Scoped diff check: `git diff --check origin/main...HEAD` -> pass.
[...]

PR #517, opened 2026-07-03
perf(grid): virtualize grid rendering

JustYannicc:
## Summary
- Virtualize Raycast Grid rendering by flattening grouped sections into positioned virtual rows and rendering only the visible/overscan window.
- Preserve hidden item/action registration, selection, context actions, keyboard navigation, scrolling, accessories, and root/section layout options (`columns`, `itemSize`, `aspectRatio`, `fit`, `inset`).
- Add a focused Grid performance/regression script for visible-cell counts, selected-item scrolling, and section layout overrides.

## Metrics
- Before runtime edits: `node scripts/test-grid-virtualization.mjs` reported `baseline eager cells=5000`, `contentResolutions=5000`, `durationMs=0.589` for a 5,000-item grouped Grid.
- After virtualization: the same 5,000-item scenario reports `virtualized visibleCells=30`, `renderedRows=7`, `totalRows=1127`, with selection-scroll coverage for item `4242`.
- Final focused run: `baseline eager cells=5000`, `contentResolutions=5000`, `durationMs=0.342`; `virtualized visibleCells=30`, `totalCells=5000`, `renderedRows=7`, `totalRows=1127`.

## Checks
- `node scripts/test-grid-virtualization.mjs` passes.
- `npm test` passes (`Renderer crash recovery (live Electron)` skipped because Electron is not launchable here).
- `npm run build:renderer` passes.
- `node scripts/check-i18n.mjs` completes successfully; it still reports the existing Italian locale gaps for `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`.
- Codex LSP diagnostics are clean for the changed Grid files.
- `./node_modules/.bin/tsc --noEmit --pretty false 2>&1 | rg "grid-runtime" || true` reports no Grid-related typecheck errors. The full renderer typecheck remains red on unrelated existing files.

## Risks / Notes
- The registration tree still mounts hidden `Grid.Item` elements so extension APIs/actions keep working; the expensive visual cell rendering and icon resolution are limited to visible rows.
- Live Electron interaction was not available in this environment, so scrolling behavior is covered by the pure layout regression rather than an end-to-end window test.

PR #516, opened 2026-07-03
perf(launcher): optimize command list rendering

JustYannicc:
## Summary
- Window large launcher command lists so only the visible row range, overscan, and selected-row neighborhood render.
- Memoize launcher row derivations for labels, icons, hotkeys, refs, and handlers with stable parent callbacks.
- Add a focused render measurement harness plus a regression test that fails if 5,000-command scenarios render every row again.

## Before / After Metrics
Measured with `node scripts/measure-launcher-command-list-render.mjs --rows=5000 --iterations=5 --warmups=1`.

| Scenario | Before row renders / median | After row renders / median |
| --- | ---: | ---: |
| root results initial | 5,000 / 123.38ms | 18 / 1.35ms |
| selection moves to middle | 5,000 / 126.65ms | 33 / 1.57ms |
| selection moves to end | 5,000 / 116.01ms | 26 / 1.33ms |
| query update same-size reshuffle | 5,000 / 129.20ms | 17 / 0.56ms |
| query update narrowed-large | 3,000 / 85.45ms | 28 / 1.17ms |
| Total scenario sequence | 23,000 / 580.69ms | 122 / 5.99ms |

## Validation
- `node --test 'scripts/test-*.mjs'` passes.
- `node --test scripts/test-launcher-command-list-windowing.mjs` passes.
- `node scripts/check-i18n.mjs` completes; existing Italian locale misses `settings.ai.whisper.vocabulary` and `settings.extensions.appSearchScope`.
- `node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:renderer` passes.
- `node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:main` passes.
- Codex LSP diagnostics for `LauncherCommandList.tsx` and `LauncherCommandRow.tsx` report no errors.

## Notes / Risks
- The virtualizer uses fixed row/header/calculator height estimates matching the current launcher row layout; future row-height changes should update those constants.
- Project-wide `./node_modules/.bin/tsc -p tsconfig.renderer.json` is currently blocked by unrelated existing errors outside this change.

PR #533, opened 2026-07-03
perf(renderer): cache icons and virtualize launcher surfaces

JustYannicc:
## What changed

- Consolidated renderer hot-path performance work for icon resolution, file icon IPC, local asset checks, emoji/list/grid virtualization, launcher command-list windowing, and stable list item registration.
- Added bounded caches/coalescing for Phosphor icon resolution, Raycast file icon data URL requests, and local asset existence checks while preserving retry behavior for misses/failures.
- Virtualized large emoji grids, Raycast grids, and launcher command-list surfaces so large result sets render only the visible window plus overscan/selection context.
- Added focused regression and measurement scripts for icon, emoji, list, grid, launcher, and native emoji caret-session paths.

## Why

These paths are all hit while rendering large launcher/Raycast result surfaces. Before this consolidation, repeated icon resolution could rescan Phosphor exports, concurrent identical file icons could fan out into duplicate IPC calls, large emoji/grid/list surfaces eagerly rendered thousands of cells/rows, and list registration could churn during selection-driven renders.

## Compatibility impact

Rendering semantics are intended to stay unchanged. Icons still resolve through the same local/remote/Phosphor/file-icon paths, file icon failures remain retryable, and local asset misses are not permanently cached so late-created files can resolve. List/grid/emoji/launcher surfaces preserve keyboard selection, scrolling, context actions, section layout options, detail rendering, and existing fallback behavior while reducing rendered DOM work.

## How tested

- `node --test scripts/test-icon-asset-cache.mjs scripts/test-icon-runtime-phosphor-cache.mjs scripts/test-icon-runtime-file-icon-cache.mjs scripts/test-emoji-grid-virtualization.mjs scripts/test-grid-virtualization.mjs scripts/test-list-registry-stable-order.mjs scripts/test-launcher-command-list-windowing.mjs scripts/test-emoji-caret-session.mjs` -> 26 passing tests.
- `pnpm test` -> 50 passing tests, 1 skipped live Electron test.
- `git diff --check` -> pass.
- Codex LSP diagnostics -> no errors for `LauncherCommandList.tsx`, `list-runtime.tsx`, `grid-runtime.tsx`, `icon-runtime-phosphor.tsx`, `icon-runtime-render.tsx`, and `grid-runtime-virtualization.ts`.
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit --pretty false` on this standalone `origin/main` branch is still blocked by existing renderer typing debt outside this consolidation, including `App.tsx`, `CameraExtension.tsx`, `LiquidGlassSurface.tsx`, launcher browser source typing, Raycast API shims, settings, snippets, and quicklink icon types.

## Performance evidence

Focused before/after measurements on a temporary `origin/main` baseline versus this branch:

| Probe | Before | After |
| --- | ---: | ---: |
| Phosphor exact/repeated, 20,000 calls | 26,845.470ms | 3.586ms |
| Phosphor unknown-fuzzy/repeated, 1,500 calls | 3,219.935ms | 3.656ms |
| Launcher scenario sequence, 5k/3k rows | 23,000 row renders / 390.968ms | 122 row renders / 4.823ms |
| Launcher root initial, 5k rows | 5,000 row renders / 81.757ms | 18 row renders / 1.184ms |

Focused tests also report emoji grid virtualization renders 112 visible emoji cells out of 4,096, grid virtualization renders 30 visible cells out of 5,000, local icon asset checks reduce 10,000 repeated positive lookups to one `statSync`, and 100 concurrent identical file icons coalesce to one IPC request.

## Stack validation

Because standalone renderer typecheck is blocked by existing `origin/main` typing debt, I also validated on a temporary stack based on upstream PR #532 (`fix(validation): restore project checks`, `JustYannicc:codex/consolidate-validation-checks` at `d4c6a9f`) plus this consolidation series. Two stack-only conflicts were resolved by keeping the validation branch type cleanups together with the list/emoji virtualization changes.

Stack results:
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit --pretty false` -> pass.
- `pnpm test` -> 51 passing tests, 1 skipped live Electron test.

## Replaces

- Replaces #484 `perf(icons): cache local asset existence checks`
- Replaces #502 `perf(list): virtualize emoji grid rendering`
- Replaces #503 `perf(icons): memoize Phosphor icon resolution`
- Replaces #507 `perf(emoji): cache caret session state`
- Replaces #515 `perf(list): stabilize item registry order`
- Replaces #516 `perf(launcher): optimize command list rendering`
- Replaces #517 `perf(grid): virtualize grid rendering`
- Replaces #526 `perf(icons): coalesce file icon IPC`

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   claude-review:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     # Optional: Filter by PR author
     # if: |
     #   github.event.pull_request.user.login == 'external-contributor' ||
@@ -41,4 +42,3 @@ jobs:
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
 
       - name: Run node test suite
         run: npm test
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "build": "npm run build:main && npm run build:renderer && npm run build:native",
     "build:main": "tsc -p tsconfig.main.json && cp src/main/emoji-data.json dist/main/emoji-data.json",
     "build:renderer": "vite build",
+    "measure:launcher-command-list": "node scripts/measure-launcher-command-list-render.mjs",
     "check:i18n": "node scripts/check-i18n.mjs",
     "test": "node --test 'scripts/test-*.mjs'",
     "build:native": "node scripts/build-native.mjs",
--- a/scripts/build-native.mjs
+++ b/scripts/build-native.mjs
@@ -28,7 +28,7 @@ const swift = [
   ['dist/native/menu-item-search', 'src/native/menu-item-search.swift',
     '-framework AppKit -framework ApplicationServices'],
   ['dist/native/emoji-trigger-monitor',
-    'src/native/emoji-trigger-monitor.swift src/native/ax-caret-query.swift',
+    'src/native/emoji-trigger-monitor.swift src/native/ax-caret-query.swift src/native/emoji-caret-session-cache.swift',
     '-framework AppKit -framework ApplicationServices'],
   ['dist/native/hotkey-hold-monitor', 'src/native/hotkey-hold-monitor.swift',
     '-framework CoreGraphics -framework AppKit -framework Carbon'],
--- a/scripts/measure-icon-resolution.mjs
+++ b/scripts/measure-icon-resolution.mjs
@@ -1,0 +1,135 @@
+#!/usr/bin/env node
+
+import { build } from 'esbuild';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const targetFile = path.join(root, 'src/renderer/src/raycast-api/icon-runtime-phosphor.tsx');
+const outputFile = path.join(os.tmpdir(), `supercmd-icon-resolution-${process.pid}-${Date.now()}.mjs`);
+
+const exactInputs = [
+  'AddPerson',
+  'Icon.ArrowLeftCircleFilled',
+  'MagnifyingGlass',
+  'Stopwatch',
+  'Temperature',
+  'Dot',
+  'XMarkCircleFilled',
+  'Folder',
+];
+
+const unknownAndFuzzyInputs = [
+  'TotallyMissingIconName',
+  'HyperSpecificSparkleClockWidget',
+  'SuperCmdTimerStopwatchShape',
+  'UnmappedTerminalCommandLineIcon',
+  'MysteryNetworkCloudBolt',
+  'Noise___No_Such_Icon_999',
+];
+
+function exposeResolverPlugin() {
+  const normalizedTarget = path.normalize(targetFile);
+  return {
+    name: 'expose-icon-resolution-for-measurement',
+    setup(buildApi) {
+      buildApi.onLoad({ filter: /icon-runtime-phosphor\.tsx$/ }, async (args) => {
+        if (path.normalize(args.path) !== normalizedTarget) return undefined;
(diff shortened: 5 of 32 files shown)


shobhit99:
### Review: perf(renderer): cache icons and virtualize launcher surfaces

Solid, well-tested consolidation. I traced the virtualization windowing, the caches, and the launcher/grid/list selection paths; behavior looks preserved in the paths that matter, with a couple of minor things worth a look.

**Strengths**
- Layout math is extracted into pure, unit-tested helpers (`grid-runtime-virtualization.ts`, `list-runtime-hooks.ts` additions). Nice separation.
- All three virtualizers keep the **selected entry mounted even when scrolled out of the window** (`getVirtualizedEntries` unions a selected-range; grid uses `getScrollTopForItemIndex`; list maps item→row). So keyboard nav to off-screen items and its `itemRefs` registration still work — the main correctness risk in windowing is handled.
- Icon-resolution caches are keyed purely on the input **name**; tint/color/theme are applied later in `renderPhosphorIcon`, not during resolution, so the cache keys are correct and theme/tint-safe. File-icon requests coalesce via an in-flight map, and **negative results are not permanently cached** (IPC errors + local-asset misses stay retryable) as claimed. Caches are bounded (clear-on-overflow / LRU).
- Launcher virtualization is gated behind a 120-entry threshold, so the common small-list case still renders through the original path (now fragment-wrapped, but DOM-equivalent under `space-y-0.5`).
- List registry order stabilized via a mount-time `orderRef` (matches the existing grid pattern) — removes per-render re-registration churn. Grid column clamp (1–8) and `itemSize`→columns mapping now match Raycast semantics; `GridInset` value change (`small`→`sm`…) is absorbed by `normalizeGridInset`, which accepts both forms.

**Concerns / potential bugs**
- **Redundant/competing scroll-into-view in the launcher (medium).** `App.tsx` `scrollToSelected` (lines 1634–1652) still calls `selectedElement.scrollIntoView({ behavior: "smooth" })` on every `selectedIndex` change, and the new `LauncherCommandList` selected-entry effect *also* calls `element.scrollTo({ behavior: "smooth" })` on the same container whenever `shouldVirtualize`. Both fire per keypress and target the same element/position; they mostly agree, but two smooth scrolls per keystroke are redundant and can visibly fight/stutter. Not a correctness break (the selected node is always rendered), but consider having one owner — e.g. gate the in-component effect off, or skip App`s scroll when the list virtualizes.
- **Launcher virtual row heights are hardcoded estimates (low).** `COMMAND_ROW_HEIGHT=38` / `SECTION_HEADER_HEIGHT=26` / `CALCULATOR_CARD_HEIGHT=124` drive absolute positioning. Rows are `px-3 py-2` around a 20px icon; if real height diverges (e.g. a wrapping subtitle) absolutely-positioned rows can overlap or leave gaps. Selection stays correct because App`s scroll uses real `getBoundingClientRect`, but visuals could drift. The list/emoji-grid paths pin exact heights inline so they are drift-free — only the launcher`s absolute layout relies on estimates. Worth confirming the constants against the real DOM.
- **Positive asset-existence cache never invalidates on deletion (low).** `positiveLocalPathExistsCache` (`icon-runtime-assets.tsx`) caches `exists=true` until a size-based flush. Misses are correctly not cached (late-created files resolve), but a deleted extension asset keeps returning `true` → a broken `sc-asset://` reference. Low real-world risk since extension asset paths are stable; noting for completeness.

**Notes**
- `itemSize="small"` grids with no explicit `columns` now render 8 columns instead of the old fixed 5 (intentional, matches Raycast). Fine, just a visible change for those extensions.
- No user-facing strings added/removed, so no i18n locale work needed. Imports/types for the touched files check out (`ListItemAccessory` imported, grid-runtime hooks imported, `data-idx` querySelector consumer replaced by index math).

🟡 Approve with minor comments — no behavior-breaking bugs found; the scroll-effect redundancy is the one thing I would resolve before merge.

JustYannicc:
Addressed the review feedback in 513f43c.

Fixes:
- Removed the virtualized launcher list selected-entry `scrollTo({ behavior: "smooth" })` effect, leaving selected-row smooth scrolling owned by `App.tsx` so there is no competing double scroll per keypress.
- Pinned virtualized launcher row/header/calculator rendered heights to the same constants reserved by the absolute layout, and extended the launcher windowing measurement/test coverage to assert those fixed heights plus the no-child-smooth-scroll contract.
- Changed the positive extension asset existence cache to refresh after a TTL, so deleted local assets stop resolving instead of staying cached until a size flush. Added a deletion/TTL regression test.

Verification run locally:
- `node --test scripts/test-launcher-command-list-windowing.mjs`
- `node --test scripts/test-icon-asset-cache.mjs`
- `node --test scripts/test-icon-runtime-phosphor-cache.mjs`
- `node --test scripts/test-icon-runtime-file-icon-cache.mjs`
- `node --test scripts/test-grid-virtualization.mjs`
- `node --test scripts/test-emoji-grid-virtualization.mjs`
- `node --test scripts/test-list-registry-stable-order.mjs`

Also confirmed via GraphQL that PR #533 has no inline review threads; the actionable feedback was the top-level review comment.

PR #531, opened 2026-07-03
perf(ai): batch streaming and status updates

JustYannicc:
## What changed
- Batches visible AI chat stream updates through a small stream buffer and forces final flushes before completion, error persistence, cancellation, and mode transitions.
- Batches Cursor Prompt streamed result rendering while keeping the authoritative result ref current for apply/finalization.
- Batches Raycast compatibility `useAI` streaming data updates with frame/timer scheduling and synchronous terminal flushes.
- Coalesces Whisper.cpp, Parakeet, and Qwen3 AI model status polling into one poll tick with semantic equality guards before React state updates.

## Why
Streaming AI text and settings model status polling previously caused renderer-visible state updates once per token/chunk or once per individual status poll. That created avoidable React work during long responses and while the AI settings tab sat open. The new helpers preserve authoritative data immediately while reducing visible updates to flush cadence or actual semantic status changes.

## Compatibility impact
No IPC contracts, public hook return shapes, user-facing strings, or locale files changed. Chat and Cursor Prompt still flush final content before persistence/apply, stream errors still surface with the accumulated text, cancellation/abort paths clear pending repaint work, `useAI` still calls `onData` with the final full text, and model download progress still updates because progress fields participate in equality checks.

## How tested
- `node scripts/test-ai-chat-stream-batching.mjs` passed; baseline 240 chunks -> 240 visible updates / 240 layout measurements / 267,795 simulated markdown chars, batched burst -> 1 visible update / 1 layout measurement / 2,290 simulated markdown chars.
- `node --test scripts/test-cursor-prompt-stream-batching.mjs` passed; baseline 500 chunks -> 500 visible result updates, batched -> 1 visible update, plus final flush/apply-before-visible-flush/cancellation coverage.
- `node --test scripts/test-use-ai-stream-batching.mjs` passed; baseline 600 chunks -> 600 visible updates, batched -> 1 visible update, real hook harness -> 3 renders / 1 visible data update, plus final data/error/abort coverage.
- `node --test scripts/test-ai-model-status-polling.mjs` passed; unchanged 5-tick baseline -> 15 state updates / 15 commits, coalesced unchanged -> 0 updates / 0 commits, all-changing -> 5 updates / 5 commits.
- `pnpm test` passed on this standalone branch: 40 pass / 1 skipped live Electron test.
- `pnpm run check:i18n` completed on this standalone branch; it still reports existing Italian locale gaps unrelated to this PR.
- `git diff --check origin/main..HEAD` passed.
[...]

PR #539, opened 2026-07-03
perf(utils): ignore stale useFetch responses

JustYannicc:
## What changed

- Add a latest-run guard to `useFetch` so older in-flight `window.electron.httpRequest` responses cannot update data, errors, loading state, or callbacks after a newer request starts.
- Invalidate active runs when `execute` flips false or the hook unmounts.
- Refetch function-valued URLs when their resolved initial request URL changes, while skipping duplicate internal rerenders for the same resolved URL/options.
- Add a focused hook harness covering stale `revalidate` responses and function URL request-input changes.

## Why

Rapid `revalidate`, pagination, or request-input changes can leave multiple HTTP requests in flight. Before this patch, an older response could resolve after a newer run and overwrite newer data or invoke stale callbacks.

## Compatibility impact

- `useFetch` return shape and option names are unchanged.
- No abort signal is added to the Electron HTTP request path; stale responses are ignored when they settle.
- Function URL behavior remains conservative: a fresh function identity does not refetch by itself if it resolves to the same initial request URL with the same request options.

## How tested

- Before evidence: `node --test scripts/test-use-fetch-stale-runs.mjs` failed before the hook change because stale data overwrote the latest result, and a changed function URL request input did not start a second request.
- After evidence: `node --test scripts/test-use-fetch-stale-runs.mjs` passes on this PR branch.
- Stack validation from the integration branch: `node --test scripts/test-use-fetch-stale-runs.mjs scripts/test-abortable-promise-hooks.mjs scripts/test-menubar-storage-remounts.mjs scripts/test-raycast-cache-size-index.mjs` passed 20 tests.
- Clean PR branch note: the nearby stack test scripts are not present on current `origin/main` yet, so the focused `useFetch` test was rerun here.
- LSP diagnostics: no TypeScript errors reported for `src/renderer/src/raycast-api/hooks/use-fetch.ts`.

## Stack validation

Draft until the consolidation stack lands. This targets upstream `main` and depends on #530, #531, #532, #533, #534, and #535.

### DIFF
--- a/scripts/test-use-fetch-stale-runs.mjs
+++ b/scripts/test-use-fetch-stale-runs.mjs
@@ -1,0 +1,322 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+let activeHost = null;
+
+const fakeReact = {
+  useCallback: (...args) => activeHost.useCallback(...args),
+  useEffect: (...args) => activeHost.useEffect(...args),
+  useMemo: (...args) => activeHost.useMemo(...args),
+  useRef: (...args) => activeHost.useRef(...args),
+  useState: (...args) => activeHost.useState(...args),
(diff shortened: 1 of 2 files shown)


JustYannicc:
Superseded by consolidated upstream PR #555.

PR #549, opened 2026-07-03
perf(icons): cache transient missing asset paths

JustYannicc:
## What changed

- Added a bounded 250ms miss cache around renderer icon asset path existence checks before calling synchronous `window.electron.statSync`.
- Kept existing asset URL semantics intact for `sc-asset://`, absolute paths, and extension-relative image filenames.
- Added a focused icon asset harness covering existing asset resolution, late-available assets after the stale window, configured `assetsPath`, and repeated miss measurement.

## Why

Repeated missing icon asset paths can synchronously cross renderer-to-main IPC on every render. The short stale window collapses bursty repeated misses without permanently hiding assets that appear after an initial miss.

## Compatibility impact

Misses can remain stale for up to 250ms. After that window, the next resolution checks the file system again, so newly available assets become visible. The miss cache is bounded to 4096 entries and clears when full.

## Metrics

- Before on the integration worker base: `iterations=10000 statSync=10000 elapsedMs=3.553` for repeated missing relative asset resolution.
- After on this clean PR branch: `iterations=10000 statSync=1 elapsedMs=5.201` for the same repeated missing relative asset resolution.
- Elapsed timings use the cheap local test stub; the representative signal is the synchronous IPC call count reduction.

## How tested

- `node scripts/test-icon-asset-cache.mjs`
- LSP diagnostics on `src/renderer/src/raycast-api/icon-runtime-assets.tsx`: no errors
- `node scripts/test-icon-runtime-file-icon-cache.mjs` passed on the integration worker branch; that script is not present on `origin/main` for this clean PR branch
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit` currently fails on unrelated upstream renderer type errors outside this change

## Stack validation

Clean branch from `origin/main`; only this scoped fix is included.

### DIFF
--- a/scripts/test-icon-asset-cache.mjs
+++ b/scripts/test-icon-asset-cache.mjs
@@ -1,0 +1,169 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+import { performance } from 'node:perf_hooks';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+
+function createIconRuntimeHarness() {
+  const moduleCache = new Map();
+  const existingPaths = new Set();
+  let statCalls = 0;
+  let nowMs = 0;
+
+  const windowStub = {
+    electron: {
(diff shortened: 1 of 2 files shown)


JustYannicc:
Superseded by consolidated upstream PR #555.

PR #552, opened 2026-07-03
perf(files): map launcher file result commands by path

JustYannicc:
## What changed

- Builds a first-match `Map` from launcher file result paths to their generated commands.
- Uses that map when assembling root-search file candidates instead of scanning `fileResultCommands` with `.find` for each result.
- Extracts the file-candidate assembly into a focused utility and adds a 3000-result `node --test` harness.

## Why

Launcher file candidate assembly can receive up to 3000 file results. Looking up each result command with `.find` makes that path O(n^2), while a path map keeps the assembly O(n) and preserves the existing path matching behavior.

## Compatibility impact

No launcher behavior is removed. File result order, strict path matching, command data, scoring, and missing-command fallback behavior are preserved. Duplicate command paths continue to use the first command, matching the old `.find` semantics.

## How tested

- Baseline lookup micro-benchmark before the change, 3000 results/commands, 500 iterations: `.find` median 12.214ms (min 11.199, max 55.398), path map median 0.100ms (min 0.083, max 0.656), equal checksum 11445000.
- After lookup micro-benchmark using the exported map builder, 3000 results/commands, 500 iterations: `.find` median 18.121ms (min 17.277, max 31.698), path map median 0.024ms (min 0.023, max 0.324), equal checksum 11445000.
- `node --test scripts/test-launcher-file-result-command-map.mjs` on the clean PR branch: full candidate assembly map median 25.097ms vs find-backed median 45.257ms, equal checksum 99347250.
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false` was run on the clean PR branch and fails on existing `origin/main` renderer type errors unrelated to this file-result lookup change.
- `git diff origin/main --check`

## Stack validation

Clean branch from `origin/main`; this PR is one commit ahead and is not stacked on #530-#535.

### DIFF
--- a/scripts/test-launcher-file-result-command-map.mjs
+++ b/scripts/test-launcher-file-result-command-map.mjs
@@ -1,0 +1,215 @@
+#!/usr/bin/env node
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { build } from 'esbuild';
+
+const root = path.resolve('.');
+const {
+  buildLauncherFileCandidates,
+  buildLauncherFileResultCommandByPath,
+} = await importBundledTs(path.join(root, 'src/renderer/src/utils/launcher-file-candidates.ts'));
+
+async function importBundledTs(absPath) {
+  const result = await build({
+    absWorkingDir: root,
+    entryPoints: [absPath],
+    bundle: true,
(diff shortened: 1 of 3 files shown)


JustYannicc:
Superseded by #555, which now includes this launcher file-result command map fix and preserves the first duplicate-path command semantics/evidence.

PR #551, opened 2026-07-03
fix(raycast): guard stale promise hook runs

JustYannicc:
## What changed

- Added an internal latest-run id to `usePromise` and `useCachedPromise` so stale completions cannot update `data`, `error`, `isLoading`, `onData`, or `onError` when callers do not pass `abortable`.
- Kept the existing AbortController guard in place for abortable callers.
- Extended the focused promise hook harness with default non-abortable overlap coverage for stale success, loading, and error paths.

## Why

Default Raycast-style hook usage can overlap executions through `revalidate()` or argument changes without passing `abortable`. The previous current-run check treated every non-abortable run as current, so an older completion could overwrite a newer result or callback state.

## Compatibility impact

- Hook return shapes and options are unchanged.
- Abortable callers still receive an `AbortController`.
- Starting a newer abortable run still aborts the previous run and clears refs on completion/unmount.

## How tested

- Before the fix, the new default overlap tests failed in `node --test scripts/test-abortable-promise-hooks.mjs`: both default `usePromise` and `useCachedPromise` let stale data overwrite `latest`.
- `node --test scripts/test-abortable-promise-hooks.mjs` now passes 12/12.
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false`
- LSP diagnostics on both changed hook files: no errors.

## Stack validation

This is a stacked draft PR. It depends on #530, #531, #532, #533, #534, #535, and `JustYannicc:codex/perf-consolidated-integration-stack`.

A clean `codex/perf-promise-latest-run-guards-pr` branch from `origin/main` was attempted, but cherry-picking this scoped commit conflicted because `origin/main` does not yet contain the abortable hook lifecycle changes or `scripts/test-abortable-promise-hooks.mjs`.

PR #548, opened 2026-07-03
perf(raycast): cache registry dirty signatures

JustYannicc:
## What changed
- Replaced List/Grid full-registry dirty snapshots with per-item visible signature caches.
- Added focused regression/metrics coverage for recreated-but-equivalent props and visible field changes.
- Kept the clean branch scoped to current `origin/main` runtime shapes.

## Why
Large registries were still rebuilding snapshot strings for unchanged recreated item props before virtualization could help. This lets unchanged item updates avoid full-registry walks and skip registry version publishes.

## Compatibility impact
- List signatures cover id, order, section, title, subtitle, icon, accessories, keywords, detail, quickLook, and action type.
- Grid signatures cover id, order, section title, title, subtitle, content, accessory, keywords, quickLook, and action type.
- Ordering, selection, selected actions, details, filtering keywords, emoji/icon values, and Raycast API-facing item fields remain represented in dirty checks.

## How tested
- `node scripts/test-registry-dirty-checks.mjs`
- `node scripts/test-registry-dirty-checks.mjs --report`
- LSP diagnostics on touched List/Grid hook files: no errors
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit` reaches TypeScript but fails on existing `origin/main` errors outside this PR, including `App.tsx`, `CameraExtension.tsx`, launcher command model/control types, i18n runtime, and existing raycast-api/settings utility issues.

## Stack validation
- Clean PR branch: `codex/perf-registry-dirty-checks-pr` from `origin/main` with only this scoped fix.
- Worker branch: `codex/perf-registry-dirty-checks` from `fork/codex/perf-consolidated-integration-stack`; validated there with stable-order, emoji-grid virtualization, and grid virtualization scripts before clean-branch adaptation.

PR #546, opened 2026-07-03
perf(icons): cache tint color resolution

JustYannicc:
## What changed
- Added bounded caches for CSS tint normalization, validation, parsed RGB values, root CSS variable reads, and readable tint results.
- Added theme/style invalidation so dark/light switches and root style changes do not leave stale readable colors.
- Added a focused tint cache regression/measurement harness.

## Why
Icon tint and readable-color helpers run during list/grid/accessory rendering. Repeated DOM element creation and computed-style reads are avoidable for repeated static tint colors.

## Compatibility impact
- Preserves named CSS colors, hex normalization, invalid-color handling, object light/dark tint selection, theme changes, and minimum-contrast adjustment.
- Dynamic CSS expressions such as var(...) are not cached as parsed/readable values, so they keep recomputing like the previous behavior.

## Metrics
10k repeated calls on the focused harness:
- resolveTintColor string: 10,000 element creations before -> 1 after.
- object light/dark tint: 20,000 element creations before -> 2 after.
- resolveReadableTintColor: 20,000 element creations and 20,000 getComputedStyle calls before -> 2 and 2 after.

## How tested
- node scripts/test-icon-tint-cache.mjs
- CLANG_MODULE_CACHE_PATH=/private/tmp/supercmd-clang-module-cache node --test 'scripts/test-*.mjs'
- LSP diagnostics on edited files: no errors

## Stack validation
Clean PR branch created from origin/main and cherry-picked only this fix. No integration-stack dependency.

PR #544, opened 2026-07-03
perf(grid): binary-search visible virtual rows

JustYannicc:
## What changed

- Replaced Grid visible-row filtering with binary bounds over row bottoms/tops, then slices the matching row range.
- Added focused tests for inclusive viewport/overscan boundaries and an instrumented 100k-row lookup shape check.

## Why

`getVisibleVirtualRows` previously scanned every virtual row on each viewport update. The new lookup keeps scroll updates tied to logarithmic bound checks plus the visible slice, which avoids total-row scaling for large grids.

## Compatibility impact

- Preserves the existing overlap predicate: `row.top + row.height >= start && row.top <= end`.
- Keeps partially visible rows and overscan edge inclusions intact.
- No public API changes.

## How tested

- Before metrics, linear scan over 1,000 viewport queries:
  - 5,000 items / 1,001 rows: 6.282 ms
  - 50,000 items / 10,001 rows: 60.661 ms
  - 100,000 items / 20,001 rows: 100.916 ms
- After metrics, binary range lookup over the same 1,000 viewport queries:
  - 5,000 items / 1,001 rows: 1.060 ms
  - 50,000 items / 10,001 rows: 0.458 ms
  - 100,000 items / 20,001 rows: 0.781 ms
- `node scripts/test-grid-virtualization.mjs`
- `pnpm exec tsc -p tsconfig.renderer.json --noEmit`
- `pnpm test` (163 passed, 1 skipped)
- TypeScript LSP diagnostics on `src/renderer/src/raycast-api/grid-runtime-virtualization.ts`: no errors; references are declaration plus the `grid-runtime.tsx` import/use.

## Stack validation

Draft until the consolidated performance stack lands. Depends on consolidated PRs #530-#535 and integration branch `JustYannicc:codex/perf-consolidated-integration-stack`.

This branch is intentionally stacked on that integration branch: the virtualization helper and focused test script are introduced by the dependency stack, so a clean `origin/main` cherry-pick cannot represent only this fix yet.

PR #555, opened 2026-07-03
perf(raycast): reduce renderer hook and registry churn

JustYannicc:
## What changed
- Ignore stale `useFetch` responses and stale promise hook completions across `usePromise`, `useCachedPromise`, and `useStreamJSON`.
- Abort and ignore stale `useStreamJSON` lifecycle work on revalidation/unmount while composing caller-provided abort signals without mutating them.
- Skip no-op form error state updates when `setValue`/`setError` would publish the same error state.
- Use binary search for visible grid virtual row bounds while preserving section layout options.
- Cache Raycast list/grid registry visible signatures to avoid full dirty-snapshot walks on equivalent re-renders.
- Cache icon tint color resolution, readable tint DOM parsing, positive local icon asset paths, and transient missing asset paths.
- Map launcher file-result commands by path for root-search candidate assembly while preserving first duplicate-path command semantics and result ordering.
- Align the consolidated grid registry dirty-check harness with the integration stack's section object model.

## Why
These renderer hot paths are hit during extension renders, hover/selection churn, virtualized grid scrolling, icon-heavy surfaces, JSON streaming hooks, and launcher file search. The consolidation keeps stale async completions from overwriting newer state and reduces repeated DOM, filesystem, registry, and launcher lookup work without changing the Raycast API surface.

## Compatibility impact
Public Raycast hook, form, grid, list, icon, and launcher APIs are unchanged. Stale async runs are ignored only when a newer run has superseded them or the hook is unmounted. `useStreamJSON` observes caller-provided abort signals via a composed signal and does not abort caller-owned controllers. Missing local asset paths are cached only for a short stale window so newly-created assets can still resolve shortly after appearing. Launcher duplicate file-result paths continue to use the first command, matching the old `.find` semantics.

## How tested
- `pnpm exec node --test scripts/test-use-fetch-stale-runs.mjs scripts/test-form-runtime-error-state-noop.mjs scripts/test-grid-virtualization.mjs scripts/test-icon-tint-cache.mjs scripts/test-registry-dirty-checks.mjs scripts/test-icon-asset-cache.mjs scripts/test-abortable-promise-hooks.mjs scripts/test-launcher-file-result-command-map.mjs scripts/test-use-stream-json-lifecycle.mjs`
- `pnpm exec vite build`
- `git diff --check fork/codex/consolidate-raycast-renderer-hooks..HEAD`
- Codex LSP focused diagnostics on touched renderer TS/TSX files: `use-fetch.ts`, `use-promise.ts`, `use-cached-promise.ts`, `use-stream-json.ts`, `useLauncherCommandModel.ts`, `launcher-file-candidates.ts`, `grid-runtime-hooks.ts`, `grid-runtime-virtualization.ts`, `list-runtime-hooks.ts`, `icon-runtime-assets.tsx`, `form-runtime.tsx`, and `form-runtime-state.ts`.
- Codex LSP references checked for the launcher file command map/candidate helpers and list/grid visible-signature builders.

## Performance evidence
- Grid virtualization harness: 5,000 eager cells before vs 30 visible cells after; 100,000-row visible range rendered 55 rows with 51 bound reads.
- Icon asset cache harness: 10,000 repeated positive path resolutions used 1 `statSync` call; 10,000 repeated missing path resolutions within the stale window used 1 `statSync` call.
- Icon tint cache harness: 10,000 repeated string tint resolutions used 1 element creation; readable tint repeat used 2 element creations, 1 body append, and 2 `getComputedStyle` calls.
- Registry dirty-check harness: single recreated equivalent list/grid item avoided full 5,000-item snapshot walks and built 1 visible signature; recreated equivalent full sets avoided full snapshot walks and produced 0 version updates.
- Launcher file candidate harness: 3,000-result assembly preserved first duplicate-path command/order and measured path-map median 26.381ms vs find-backed median 46.192ms with equal checksums.
- Promise/fetch/stream stale-run harnesses cover latest-run wins for stale data and stale errors across default, abortable, revalidation, and unmount paths.
- `useStreamJSON` before evidence from #554: the lifecycle harness failed before the hook change for stale data overwrite, stale error surfacing, missing unmount cleanup, and missing signal composition; it passes after this consolidation.

## Stack validation
This branch starts from `codex/perf-consolidated-integration-stack` and adds only the Raycast renderer hook, registry, grid, form, icon, launcher file lookup, and `useStreamJSON` lifecycle fixes assigned here. Renderer typecheck and production renderer build pass on this stack.

## Replaces
- #539
- #542
- #544
- #546
- #548
- #549
- #551
- #552
- #554


shobhit99:
Reviewed the Raycast-runtime portion of this stack (the 10 commits on top of `codex/perf-consolidated-integration-stack`): the promise/fetch/stream hooks, list/grid registries, grid virtualization, icon asset/tint caches, form error state, and launcher file-candidate extraction. No public Raycast API surface changes; hook return shapes and `isLoading/data/error/mutate/revalidate` semantics are preserved. This PR's own diff carries no user-facing strings (i18n N/A).

### Strengths
- **Stale-run guarding is the real win.** The monotonic `latestRunIdRef`/`runIdRef` guards in `usePromise`, `useCachedPromise`, `useFetch`, and `useStreamJSON` correctly make "latest run wins." For the **non-abortable** path this fixes a genuine bug: previously `isCurrentRun = () => !runController || ...` returned `true` for every concurrent run, so an out-of-order-resolving stale run could clobber newer state. The added `latestRunIdRef.current === runId` term is essential there and redundant-but-harmless for the abortable path (each `execute` reassigns `abortControllerRef` in lock-step). Sequential pagination in `useCachedPromise`/`useFetch` is unaffected (the appending page is always the latest run).
- **`useStreamJSON` abort composition is well done.** `composeAbortSignal` observes the lifecycle signal + the caller's `signal` and aborts only its *own* controller, never the caller's — matching the PR claim. It correctly distinguishes caller-abort (surfaces the error, as before) from lifecycle-abort (`lifecycleController.signal.aborted` / superseded `runId` → suppressed), removes listeners in `finally`, and only nulls `abortControllerRef` when it still owns it, so a late-resolving superseded run can't clobber the new run's ref or `isLoading`.
- **Registry churn reduction is correct and even fixes a gap.** Per-item visible signatures replace the O(n) full-snapshot walk on every microtask. The list/grid signatures are a *superset* of the old snapshot fields — grid now additionally tracks `content`/`subtitle`/`accessory`/`keywords`/`quickLook`, which the old grid snapshot omitted — so there are no missed version bumps (no stale rows), and grid content changes are now detected. The early-returns fire only on provably no-visible-change. Critically, because the integration base assigns `order` via a stable `orderRef` (assigned once, dropped from the `useLayoutEffect` deps in `list-runtime-renderers.tsx` / `grid-runtime-items.tsx`), `registry.set` is *not* invoked on selection/hover re-renders, so adding `orderChanged`/`String(order)` to the signature introduces no per-keystroke churn.
- **Grid virtualization binary search** (`grid-runtime-virtualization.ts`) is correct for a monotonic row layout: `slice(firstBottom≥start, firstTop>end)` reproduces the old filter's `top+height >= start && top <= end` boundaries exactly, including empty-range edges.
- **Icon caches** key theme-dependent color/tint results by a theme signature + version (invalidated via a `MutationObserver` on `documentElement` `class`/`style`), cache theme-independent normalization/validity separately, and correctly *exclude* `var()/env()/currentcolor/inherit/...` from caching. Negative asset-path cache uses a 250 ms TTL so freshly-created assets still resolve.
- **`launcher-file-candidates.ts`** replaces a per-result O(n) `.find(item.path === result.path)` with a prebuilt first-wins `Map`, preserving both the first-duplicate-wins semantics and result ordering; otherwise a faithful extraction. **Form** no-op skips return the same reference (React bails the re-render) and skip the redundant `setCurrentFormErrors` publish while the global stays in sync.

### Concerns / potential bugs
- **[Low] `use-fetch.ts` — function-URL effect keying.** The init effect dep changed from a constant `urlString` (`'function'` for function URLs) to `url`, with dedup on the *resolved* page-0 URL via `lastInitialRequestKeyRef`. This is a net improvement — it fixes a real staleness bug where a function URL embedding changing state (e.g. `() => \`...?q=${query}\``) never refetched under the old constant key. The edge risk: a function URL that resolves *non-deterministically* (embeds `Date.now()` / a random nonce) now yields a new `initialRequestKey` every render, so `fetchData(0)` fires each render (each fetch toggles `isLoading` → re-render → refetch), i.e. a fetch loop. This pattern is unusual/arguably already broken, but it's a behavior change worth a note.
- **[Very low] `icon-runtime-assets.tsx` readable-tint cache.** Theme detection keys off the `documentElement` `dark` class + class/style signature. If a theme CSS var like `--surface-base-rgb` ever changed without a corresponding class/style mutation on `documentElement` (e.g. a stylesheet swap), a cached readable tint could be briefly stale. Low risk given the app toggles theme via the `dark` class; cosmetic only.
- **[Nit] `parseCssColorToRgb`** removes the temp `<span>` inside the `try` after `getComputedStyle`; if `getComputedStyle` itself threw, the span would leak. Practically impossible, and still strictly better than the previous un-guarded version.

### Suggestions
- Consider whether `use-fetch` should snapshot `typeof url === 'function'` and only re-key on the resolved URL when the function identity *and* resolved value both change, to fully neutralize the non-deterministic-URL edge — or just document the expectation that function URLs resolve deterministically.

🟡 Approve with minor comments — behavior/API is preserved and the stale-run and abort fixes are genuine improvements; only low-severity edge cases noted.

JustYannicc:
Addressed the review concerns in `1fd471e`:

- Hardened `useFetch` function URL keying so hook-owned state renders do not re-resolve a non-deterministic URL function into a fetch loop, while caller rerenders with a changed resolved URL still refetch.
- Extended readable tint cache invalidation to observe stylesheet/head changes in addition to `documentElement` class/style changes.
- Moved temporary CSS color parse element cleanup into `finally` so a thrown `getComputedStyle` cannot leak the node.
- Added regression coverage for the non-deterministic URL case, stylesheet-driven tint invalidation, and temporary element cleanup on parser failure.

Proof run locally:

- `node scripts/test-use-fetch-stale-runs.mjs`
- `node scripts/test-icon-tint-cache.mjs`
- `node scripts/test-registry-dirty-checks.mjs`
- `node scripts/test-abortable-promise-hooks.mjs`
- `node scripts/test-use-stream-json-lifecycle.mjs`
- `node scripts/test-icon-asset-cache.mjs`
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit`

PR #567, opened 2026-07-04
perf(emoji): reuse inline picker matches while navigating

JustYannicc:
## What changed
- Added a module-level cache for the currently rendered inline emoji picker matches.
- Render refreshes the cache for each new query; hide clears it.
- Left/right navigation and enter/tab insertion now reuse the rendered match list instead of rescanning emoji data.
- Added a VM-based Node test harness for the inline emoji picker path without importing the full Electron main entrypoint.

## Why
- The inline picker query does not change during arrow navigation, but the previous path searched emoji data in `handleEmojiTriggerNav` and again in `updateEmojiPickerSelection` for each left/right key.
- Reusing the rendered matches removes those repeated full scans while keeping the same ordered result list.

## Compatibility impact
- Emoji ordering remains the static `emoji-data.json` search ordering produced during render.
- Excluded-app dismissal, picker hide/intercept behavior, zero-match hide, selection wraparound, enter/tab insertion, and escape dismissal are preserved.
- Cached matches are cleared on hide so stale query data is not reused after dismissal or insertion.

## How tested
- `node --test scripts/test-inline-emoji-picker-match-cache.mjs`
- `node --test 'scripts/test-*.mjs'` (238 pass, 1 skip, 0 fail)
- `./node_modules/.bin/tsc -p tsconfig.main.json --pretty false`
- Codex LSP diagnostics for `src/main/main.ts`: no error diagnostics; all-severity check showed only preexisting hint-level items outside this change.

## Performance evidence
- Harness data: `src/main/emoji-data.json` with 1,870 entries, query `smi`, 500 repeated sequences of 20 right-arrow navigation steps.
- Legacy repeated-navigation search model: 20,000 search calls, 2,899.988 ms total, 5.8 ms per sequence.
- Cached model: 500 search calls, 76.143 ms total, 0.1523 ms per sequence.
- Search-call reduction: 40x, with matching checksum `113000`; the VM path also asserts left/right/enter perform 0 additional searches after render.

## Stack validation
- Started from `fork/codex/perf-consolidated-integration-stack` at `264fb89dc5c867d6c3d77bc4b95e26e235e326ea`.
- Work branch: `codex/perf-emoji-picker-match-cache`.
- Commit: `2cbcf35` (`perf(emoji): reuse inline picker matches while navigating`).

## Replaces
- None.


JustYannicc:
Superseded by the consolidated wave-2 UI/runtime PR: https://github.com/SuperCmdLabs/SuperCmd/pull/572

PR #571, opened 2026-07-04
perf(commands): avoid rediscovery for metadata updates

JustYannicc:
## What changed
- Added a command metadata patch path that updates cached runtime subtitles by command id, including Raycast extension path ids such as `extension/command`.
- Replaced subtitle-only `invalidateCache()` calls for inline script completion and `update-command-metadata` with the metadata cache patch helper.
- Kept the command disk cache structural by saving base subtitles and reapplying current runtime metadata on cache load.
- Added a focused metadata-cache harness that compares the legacy invalidation path with the optimized patch path.

## Why
Runtime subtitle updates were persisted into `commandMetadata` and then invalidated the whole command cache. The next `getAvailableCommands()` could therefore start full app/settings/extension/script/quick-link discovery even though only one command subtitle changed.

## Compatibility impact
Settings format is unchanged. Aliases, installed apps, settings panes, extensions, quick links, scripts, and structural invalidation paths keep their existing behavior. Non-inline script commands still ignore runtime subtitle metadata. Raycast `updateCommandMetadata` path-style ids now apply consistently during live cache patches and command discovery.

## How tested
- `node --test scripts/test-command-metadata-cache.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit`
- Codex LSP diagnostics for `src/main/commands.ts`: no errors
- Codex LSP diagnostics for `src/main/main.ts`: no errors
- `git diff --check`

## Performance evidence
Focused harness output for 5,000 cached commands over 60 metadata updates:
- Legacy invalidation path: `structuralDiscoveryStarts=60`, median `getCommandsMs=0.115`
- Optimized metadata patch path: `structuralDiscoveryStarts=0`, median `getCommandsMs=0.014`
- Avoided structural discovery starts: `60`

The harness uses a fake discovery runner so the measurement counts discovery starts deterministically without scanning the local machine.

## Stack validation
Started from `codex/perf-consolidated-integration-stack` / `fork/codex/perf-consolidated-integration-stack` at `264fb89dc5c867d6c3d77bc4b95e26e235e326ea`. Implementation branch is `codex/perf-command-metadata-cache`, pushed to `JustYannicc/SuperCmd`, targeting `SuperCmdLabs/SuperCmd:main`. Dependency setup used `pnpm install --ignore-scripts --no-frozen-lockfile --lockfile=false`; no package metadata changes were staged.

## Replaces
No prior PR. Replaces the subtitle-only runtime metadata paths that broadly invalidated the structural command discovery cache.

PR #568, opened 2026-07-04
perf(commands): cache script command icon data

JustYannicc:
## What changed
- Added a module-level file icon data URL cache for script commands, keyed by resolved absolute icon path plus the icon file size and mtime signature.
- Kept script command discovery invalidation focused on script metadata while reusing unchanged file icon data across cold discoveries.
- Extended the script command discovery benchmark with an icon-heavy fixture and invalidated discovery/execution measurements.
- Added focused coverage that verifies cached file icons avoid repeat byte reads and changed icon files refresh.

## Why
Running a script command after the 12s discovery TTL can force a cold rediscovery. File-based script icons were re-read and base64 encoded on every cold discovery, so icon-heavy command sets repeatedly paid the full icon I/O cost even when icon files had not changed.

## Compatibility impact
- Emoji icons and `https://` icon values still bypass file icon reads.
- File icon support remains limited to the existing SVG, PNG, JPG, and JPEG data URL behavior.
- Raycast metadata parsing and `iconDark`/`icon` precedence are unchanged.
- Script metadata is still refreshed on discovery cache invalidation; changed file icons are picked up when their file signature changes.

## How tested
- `node --test scripts/test-script-command-runner.mjs`
- `node --test 'scripts/test-*.mjs'` (236 pass, 1 skipped)
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit`
- Codex LSP diagnostics for `src/main/script-command-runner.ts`: no diagnostics found
- `node scripts/bench-script-command-discovery.mjs`

## Performance evidence
Command: `SUPERCMD_BENCH_SCRIPT_COUNT=250 SUPERCMD_BENCH_SCRIPT_BODY_BYTES=0 SUPERCMD_BENCH_SCRIPT_ICON_BYTES=98304 node scripts/bench-script-command-discovery.mjs`

Before cache:
- initial discovery: 31.7ms, 250 `readFileSync` calls, 23.4 MiB icon bytes
- invalidated discovery: 28.1ms, 250 `readFileSync` calls, 23.4 MiB icon bytes, 23.5 MiB total runner bytes
- invalidated execution lookup: 27.4ms, 250 `readFileSync` calls, 23.4 MiB icon bytes, 23.5 MiB total runner bytes

After cache:
- initial discovery: 33.5ms, 250 `readFileSync` calls, 23.4 MiB icon bytes
- invalidated discovery: 8.5ms, 0 `readFileSync` calls, 0 B icon bytes, 76.1 KiB total runner bytes
[...]

PR #566, opened 2026-07-04
perf(files): speed up delete tombstone batches

JustYannicc:
## What changed
- Split delete tombstoning so direct indexed file deletes mark their entry ids without scanning the whole snapshot.
- Kept indexed directory deletes and unknown deleted paths on the descendant tombstone path.
- Added focused coverage for direct-file batch behavior, unknown-path descendants, and direct/delete benchmark reporting.

## Why
Large watcher delete batches can be mostly exact indexed file paths. The previous multi-path tombstone flow scanned every indexed entry even when all deleted paths were already known files, making delete batches much slower than watcher adds.

## Compatibility impact
File-search behavior is preserved. Exact indexed file deletes are tombstoned directly, indexed directory deletes still tombstone descendants, and unknown deleted paths still tombstone indexed descendants. No feature trimming, config changes, or user-facing API changes.

## How tested
- `node --test scripts/test-file-search-delete-batch.mjs scripts/test-file-search-perf-harness.mjs` -> 7 passing tests.
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit` -> passed.
- Codex LSP diagnostics for `src/main/file-search-index.ts` -> no diagnostics found.
- `git diff --check` -> passed.

## Performance evidence
- Before, exact audit harness: `node scripts/file-search-perf-harness.mjs --projects 600 --modules 10 --files-per-module 10 --query-runs 2 --updates 500 --deletes 500 --limit 80` -> 79,804 indexed entries, watcher-style updates 7.685ms for 500 paths, delete batch 122.856ms for 500 paths.
- After, same command: 79,804 indexed entries, watcher-style updates 8.487ms for 500 paths, delete batch 5.587ms for 500 paths, deleted exact matches after tombstone 0.
- Focused benchmark after: `node scripts/benchmark-file-search-delete-batch.mjs` -> direct-file-batch 5.353ms for 500 paths over 80,161 entries; directory-tree-batch 8.561ms for 4,041 paths over 68,162 entries.

## Stack validation
Started from `264fb89dc5c867d6c3d77bc4b95e26e235e326ea` on `codex/perf-consolidated-integration-stack` / `fork/codex/perf-consolidated-integration-stack`. This branch contains one commit on top: `e4d7dd7 perf(files): speed up delete tombstone batches`. PR targets `SuperCmdLabs/SuperCmd:main` from `JustYannicc:codex/perf-file-search-delete-tombstones`.

## Replaces
None.

PR #573, opened 2026-07-04
perf(commands): reduce command and file refresh overhead

JustYannicc:
## What changed
- Batch file-search delete tombstone handling by collapsing nested delete roots and marking descendants in one pass.
- Cache script command file icon data by path signature so repeated discoveries reuse unchanged data URLs and refresh changed icons.
- Patch cached runtime command metadata in place for subtitle updates instead of invalidating command discovery.
- Add focused harnesses for delete tombstones, script command icon caching, and command metadata cache behavior on current upstream main.

## Why
These hot paths can run during filesystem watcher bursts, script command rediscovery, and @raycast/api metadata updates. The consolidated changes reduce repeated scans and rediscovery while preserving command ordering and existing search/script semantics.

## Compatibility impact
- Filesystem search semantics are preserved; deletes still tombstone exact indexed files, directory descendants, unknown deleted roots, and avoid prefix-like neighbors.
- Script command icon behavior is preserved; unchanged file icons reuse cached data, changed icon files refresh by size/mtime signature, and emoji icons are unchanged.
- Metadata updates still persist settings and notify renderers, while cached command subtitles are patched without structural rediscovery.

## How tested
- `node --test scripts/test-file-search-delete-batch.mjs`
- `node --test scripts/test-script-command-runner.mjs`
- `node --test scripts/test-command-metadata-cache.mjs`
- `SUPERCMD_BENCH_SCRIPT_ICON_BYTES=4096 node scripts/bench-script-command-discovery.mjs`
- `node scripts/benchmark-file-search-delete-batch.mjs`
- `tsc -p tsconfig.main.json --noEmit --pretty false`
- Codex LSP diagnostics for `src/main/file-search-index.ts`, `src/main/script-command-runner.ts`, `src/main/commands.ts`, and `src/main/main.ts`
- `git diff --check`

## Performance evidence
- File delete benchmark: direct-file batch tombstoned 500 of 80,161 entries in 5.681 ms; directory-tree batch tombstoned 8,041 of 68,162 entries across 4,041 delete paths in 8.145 ms.
- Script command benchmark with 40 one MiB scripts and 4 KiB file icons: initial discovery read 80 files / 40.2 MiB; invalidated discovery read 40 files / 40.0 MiB, skipping the 40 unchanged icon reads.
- Command metadata benchmark: 5,000 commands across 60 iterations avoided 60 structural discovery starts; optimized median get-commands path was 0.035 ms versus legacy 0.167 ms.

## Stack validation
- Started from clean `origin/main` at `9bdd3d2`.
- Cherry-picked the wave commits from #566, #568, and #571 onto `codex/consolidate-command-file-wave2`.
- Adjusted only focused harness assumptions needed for current upstream main; no coordinator notes, prompt files, playbooks, or package-manager metadata are included.
- Main TypeScript, LSP diagnostics, focused harnesses, benchmarks, and diff whitespace checks passed locally.

## Replaces
- #566
- #568
- #571

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   claude-review:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     # Optional: Filter by PR author
     # if: |
     #   github.event.pull_request.user.login == 'external-contributor' ||
@@ -41,4 +42,3 @@ jobs:
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
 
       - name: Run node test suite
         run: npm test
--- a/scripts/bench-script-command-discovery.mjs
+++ b/scripts/bench-script-command-discovery.mjs
@@ -1,0 +1,133 @@
+#!/usr/bin/env node
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { loadScriptCommandRunner } from './lib/script-command-runner-harness.mjs';
+
+const fileCount = Number(process.env.SUPERCMD_BENCH_SCRIPT_COUNT || 40);
+const bodyBytesPerFile = Number(process.env.SUPERCMD_BENCH_SCRIPT_BODY_BYTES || 1024 * 1024);
+const iconBytesPerFile = Number(process.env.SUPERCMD_BENCH_SCRIPT_ICON_BYTES || 0);
+
+function formatBytes(bytes) {
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KiB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MiB`;
+}
+
+function makeLargeScript(title, bodyBytes, icon = '⚡') {
+  const header = `#!/bin/bash
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title ${title}
+# @raycast.mode fullOutput
+
+# Optional parameters:
+# @raycast.packageName Benchmark
+# @raycast.icon ${icon}
+# @raycast.description Large script command fixture
+# @raycast.needsConfirmation false
+
+exit 0
+`;
+  const line = `# ${'x'.repeat(253)}\n`;
+  const repeatCount = Math.max(0, Math.ceil(bodyBytes / Buffer.byteLength(line)));
+  return header + line.repeat(repeatCount);
+}
+
+function snapshot(metrics) {
+  return {
+    readFileSyncCalls: metrics.readFileSyncCalls,
+    readFileSyncBytes: metrics.readFileSyncBytes,
+    readSyncCalls: metrics.readSyncCalls,
+    readSyncBytes: metrics.readSyncBytes,
+    openSyncCalls: metrics.openSyncCalls,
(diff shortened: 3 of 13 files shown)


shobhit99:
### Review — perf(commands): reduce command and file refresh overhead

Focused, well-tested perf refactor of three independent hot paths (command-metadata cache, file-search delete tombstones, script-command icon caching). I traced each against upstream `main`; semantics are preserved and the new harnesses are meaningful.

**Strengths**
- **Command metadata:** `applyCommandMetadataUpdate` correctly preserves the subtitle-eligibility gate (`!(script && mode !== 'inline')`), keys extension commands by both `id` and `path`, and patches the live `CommandInfo` objects shared by `cachedCommands` and `staleCommandsFallback`. Because both sites `saveSettings(...)` *before* patching, the change survives a concurrent rediscovery (which re-applies from settings via `applyStoredRuntimeCommandMetadata`). `saveCommandsDiskCache` now writes the *base* subtitle and re-applies overrides on load — correct round-trip.
- `main.ts:15240` is a net improvement: the old code only called `invalidateCache()` (never broadcast despite the "Notify all windows" comment); the new path actively `broadcastCommandsUpdated()` on change and avoids a full rediscovery.
- **File-search tombstones:** collapse is sound — sorting by length guarantees an ancestor always precedes its descendant, so `hasDeletedPathAncestor` de-dupes nested roots correctly. File-vs-directory descendant handling and the unknown-root case match the old prefix behavior. The `path.resolve` re-normalization in `normalizeDeletedPath` is safe: index entries (`file-search-index.ts:700/781/785`) and watch delete paths are already `path.resolve`'d, so keys still match `pathToEntryId`.
- **Icon cache:** `size:mtimeMs` signature, drops the entry when the file disappears, and re-reads on change — behavior parity with the old `existsSync`+read (statSync/existsSync both follow symlinks). Good targeted test for cache hit vs. refresh.

**Concerns / potential bugs**
- **Minor — `src/main/main.ts:15240` (`update-command-metadata`):** unlike the old unconditional `invalidateCache()`, the new path does nothing when `changedCommands === 0`. If a *fresh* (non-stale) cache genuinely doesn't contain the target command, the persisted subtitle won't surface in the in-memory list until `CACHE_TTL` (30 min) or the next invalidation. This self-heals in the common cases (a running extension's command is always in cache; a null `cachedCommands` re-discovers via the stale path; the value is always persisted to settings), so it's narrow — but a one-line fallback (`if (patchResult.matchedCommands === 0) invalidateCache();`) would fully restore the old guarantee.
- **Very minor — `src/main/script-command-runner.ts` (`getCachedIconDataUrl`):** a negative result (`dataUrl: undefined` from a transient read failure) is cached until the file's size/mtime changes, whereas the old code retried every discovery. Also the module-level `iconDataUrlCache` is unbounded, though bounded in practice by the number of script-command icons. Both are acceptable.

**Suggestions (optional)**
- Consider the `matchedCommands === 0 → invalidateCache()` fallback above for full parity.
- `useBackgroundRefresh.ts` is untouched (interval behavior unchanged) — worth confirming that's intentional given the PR title mentions "refresh overhead."

Overlaps conceptually with #620 (command search validation) but touches different code (discovery/refresh vs. search), so no duplication. Types are sound (`subtitle?`/`mode?`/`path?` all optional, so the in-place mutate + `delete` are valid) and the change is plausibly compiling per the reported `tsc -p tsconfig.main.json`.

🟡 Approve with minor comments

JustYannicc:
Addressed the maintainer follow-up from the top-level review.

Changes:
- Restored the old metadata-update guarantee for the narrow no-match case: `update-command-metadata` now patches the live cache when possible, but falls back to invalidating the command cache when `matchedCommands === 0`, so persisted metadata is re-applied on the next command fetch instead of waiting for the TTL.
- Added regression coverage for that fresh-cache/missing-command fallback path.
- Hardened the script-command icon cache by only caching successful data URLs, retrying after transient read failures, and capping the cache size.
- Added regression coverage for the transient icon read-failure case.

Notes:
- Confirmed via GraphQL that there are no inline review threads or formal review bodies on this PR; this addresses the actionable top-level feedback.
- `src/renderer/src/hooks/useBackgroundRefresh.ts` remains intentionally untouched: the PR changes are scoped to command metadata patching, file-search delete tombstones, and script-command icon reads; renderer interval behavior is unchanged.

Verification:
- `node --test scripts/test-command-metadata-cache.mjs scripts/test-file-search-delete-batch.mjs scripts/test-script-command-runner.mjs scripts/test-root-search-ranking.mjs` ✅ 14 passed
- `git diff --check` ✅

PR #570, opened 2026-07-04
perf(whisper): avoid unused local STT file reads

JustYannicc:
## What changed
- Extracted the file-based Whisper IPC transcription flow into a testable helper.
- Deferred `fs.readFileSync(audioPath)` until the selected STT provider requires a Buffer; `whispercpp` continues to send the captured file path directly.
- Added a focused harness covering provider read behavior, cleanup order, and instrumentation evidence.

## Why
The default local `whispercpp` path already transcribes by file path, so the previous eager read created an unused full-file Node heap allocation for native-captured audio.

## Compatibility impact
No user-visible behavior changes expected. Buffer-backed providers still read the captured WAV once and successful transcriptions still unlink the temp file before removing its directory.

## How tested
- `node --test scripts/test-whisper-file-buffer-read.mjs`
- `node --test scripts/test-whisper-file-buffer-read.mjs scripts/test-native-helper-process-lifecycle.mjs scripts/test-local-model-status-probe-cache.mjs scripts/test-ai-model-status-polling.mjs scripts/test-ai-chat-stream-batching.mjs scripts/test-ai-stream-ipc-batching.mjs scripts/test-use-ai-stream-batching.mjs scripts/test-raycast-ai-abort-listeners.mjs scripts/test-speak-settings-listener.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit`
- `git diff --cached --check`

## Performance evidence
The new harness logs an eager-read baseline of `readCalls=1 nodeBufferBytes=73400320`; after the change `whispercpp` logs `readCalls=0 nodeBufferBytes=0 avoidedNodeBufferBytes=73400320`, while `openai` logs `readCalls=1 nodeBufferBytes=73400320`.

## Stack validation
Started from `fork/codex/perf-consolidated-integration-stack` at `264fb89dc5c867d6c3d77bc4b95e26e235e326ea`. Codex LSP diagnostics on `src/main/main.ts` and `src/main/whisper-transcribe-file.ts`: no diagnostics found.

## Replaces
None

PR #565, opened 2026-07-04
perf(extensions): prune fired one-shot timers

JustYannicc:
## What changed
Wrapped extension-scoped one-shot `setTimeout` and `requestAnimationFrame` callbacks so fired handles remove themselves from the lifecycle registry and clearer maps. Extended the lifecycle sandbox harness with a flushable fake clock, explicit clear/cancel coverage, thrown-callback coverage, unmount cleanup coverage, and interval-retention assertions.

## Why
Long-lived menu-bar extensions can schedule repeated one-shot timers over hours. Those fired handles no longer need lifecycle cleanup, but they previously stayed in `registry.timeouts` and `registry.rafs`, growing memory and future cleanup work until unmount or re-evaluation.

## Compatibility impact
No extension API shape changes. Timeout callback arguments and practical browser `this` binding are preserved through the wrapper, explicit `clearTimeout` and `cancelAnimationFrame` still untrack and cancel pending handles, `setImmediate` continues to delegate through timeout cleanup, and unmount cleanup still clears pending one-shots. Interval behavior remains explicit-clear/unmount based and intervals stay tracked until cleared.

## How tested
- `node --test scripts/test-extension-lifecycle-sandbox.mjs` -> 8/8 passing
- `node --test 'scripts/test-*.mjs'` -> 242/242 passing
- Codex LSP `diagnostics` with `severity: error` on `src/renderer/src/ExtensionView.tsx` -> no diagnostics

## Performance evidence
- `1000 fired timeouts: before stale registry entries=1000, after=0`
- `1000 fired RAFs: before stale registry entries=1000, after=0`
- The same harness confirms intervals remain tracked after one-shot timeout/RAF callbacks fire, then drop to zero after explicit clear.

## Stack validation
Implementation branch was created from `fork/codex/perf-consolidated-integration-stack` after verifying both `HEAD` and that stack branch resolved to integration base `264fb89dc5c867d6c3d77bc4b95e26e235e326ea`.

## Replaces
None

PR #564, opened 2026-07-04
perf(ai): stream Gemini prompt responses

JustYannicc:
## What changed
- Switched single-prompt Gemini requests from `generateContent` to `streamGenerateContent?alt=sse`.
- Reused Gemini SSE text extraction for chat and prompt flows while keeping prompt-only no-text errors.
- Added a focused synthetic SSE regression harness for Gemini prompt streaming.

## Why
Gemini prompt calls were buffering the full provider response before yielding a single chunk, delaying first visible output for AI ask, cursor prompt, prompt-window, Raycast AI ask, and transcript correction paths.

## Compatibility impact
Final joined text and no-text error messages are preserved. The only intended user-visible change is improved provider chunk cadence and lower first-token latency for Gemini prompt responses.

## How tested
- `node --test scripts/test-ai-provider-gemini-prompt-streaming.mjs`
- `node --test scripts/test-ai-stream-ipc-batching.mjs`
- `node scripts/test-ai-chat-stream-batching.mjs`
- `node --test scripts/test-cursor-prompt-stream-batching.mjs`
- `node --test scripts/test-use-ai-stream-batching.mjs`
- `pnpm exec tsc -p tsconfig.main.json`
- Codex LSP diagnostics on `src/main/ai-provider.ts`: no diagnostics found

## Performance evidence
Synthetic Gemini SSE harness diagnostics showed legacy `generateContent` buffered 75 bytes before first yield, while `streamGenerateContent` yielded after 66 of 179 SSE bytes and before response end. The new regression also asserts incremental first-chunk delivery, final joined text equivalence, and preserved no-text errors.

## Stack validation
Started from `fork/codex/perf-consolidated-integration-stack` at `264fb89dc5c867d6c3d77bc4b95e26e235e326ea`. Implementation branch: `codex/perf-gemini-prompt-streaming`.

## Replaces
None

PR #563, opened 2026-07-04
perf(ai): avoid copying Whisper upload bodies

JustYannicc:
## What changed

- Stream Whisper/OpenAI STT multipart upload parts directly to the HTTPS request instead of concatenating them into a single body buffer.
- Keep Content-Length by summing the same multipart part lengths and preserve file/model/response_format/language ordering.

## Why

Large STT uploads previously pushed the original audio buffer into the multipart parts array and then called Buffer.concat(parts), allocating a second full multipart payload before writing to the socket.

## Compatibility impact

No user-visible behavior change. Request endpoint, headers, Content-Length, filename/content type selection, optional language field, abort/error handling, and transcript trimming are preserved.

## How tested

- Codex LSP diagnostics on src/main/ai-provider.ts: no diagnostics.
- ./node_modules/.bin/tsc -p tsconfig.main.json --noEmit
- node --test scripts/test-ai-provider-whisper-upload.mjs scripts/test-ai-stream-ipc-batching.mjs scripts/test-ai-chat-stream-batching.mjs scripts/test-use-ai-stream-batching.mjs scripts/test-raycast-ai-abort-listeners.mjs
- node --test scripts/test-cursor-prompt-stream-batching.mjs scripts/test-ai-model-status-polling.mjs

## Performance evidence

scripts/test-ai-provider-whisper-upload.mjs fakes https.request for an 8 MiB audio upload, asserts the exact multipart bytes and Content-Length, and instruments Buffer.concat during the production call. It reports that the legacy concat would allocate an 8,389,124 byte multipart body copy, while the streamed path writes 7 buffers, reuses the 8,388,608 byte audio Buffer, and allocates 516 bytes of framing.

## Stack validation

Started from fork/codex/perf-consolidated-integration-stack at 264fb89dc5c867d6c3d77bc4b95e26e235e326ea. Implementation branch: codex/perf-whisper-stream-upload. Commit: 0c60799.

## Replaces

None

PR #575, opened 2026-07-04
perf(runtime): reduce lifecycle and media churn

JustYannicc:
## What changed
- Consolidates the wave-2 lifecycle/media performance work for script command timeout cleanup, native menubar update caching, Whisper upload streaming, Gemini prompt streaming, extension one-shot timer pruning, and lazy STT file reads.
- Adds focused harness coverage for the consolidated runtime/media paths and a minimal TypeScript import test helper stub path for clean-base harnessing.

## Why
- Reduces long-running runtime churn from stale timeout handles, repeated native tray work, retained extension one-shot timers, eager STT file buffers, and avoidable upload/response buffering.
- Keeps the wave-2 fixes together on a clean `origin/main` base so upstream can review one lifecycle/media PR instead of six overlapping branches.

## Compatibility impact
- Preserves script command execution behavior, including output-limit failure handling and timeout cleanup.
- Preserves MenuBarExtra fallback title/tooltip/icon behavior while skipping unchanged native tray work.
- Preserves Whisper/STT provider routing: whisper.cpp uses file-path transcription without reading a Node buffer; buffer-backed cloud/local providers still read once and clean up temp files.
- Preserves Gemini prompt output/error semantics while using `streamGenerateContent` SSE.
- Preserves extension no-view/background lifecycle cleanup by scoping timers/listeners per extension evaluation and pruning fired one-shots.

## How tested
- `node --test scripts/test-script-command-runner.mjs scripts/test-menubar-native-update-cache.mjs scripts/test-ai-provider-whisper-upload.mjs scripts/test-ai-provider-gemini-prompt-streaming.mjs scripts/test-extension-lifecycle-sandbox.mjs scripts/test-whisper-file-buffer-read.mjs`
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false` (run because `ExtensionView.tsx` changed; fails on existing non-`ExtensionView.tsx` renderer errors in App, CameraExtension, LiquidGlassSurface, launcher hooks, raycast-api helpers, settings, and quicklink icons)
- Codex LSP diagnostics with `severity=error` for `src/main/ai-provider.ts`, `src/main/main.ts`, `src/main/menubar-native-update-cache.ts`, `src/main/script-command-runner.ts`, `src/main/whisper-transcribe-file.ts`, and `src/renderer/src/ExtensionView.tsx`: no diagnostics found
- `git diff --check`

## Performance evidence
- Script output-limit paths clear the active timeout handle after stdout/stderr overflow (`before=1 after=0 cleared=1`).
- Menubar stable title ticks reduce native menu rebuild/context-menu work from 60 calls to 1 call.
- Whisper multipart upload writes 7 buffers and reuses the 8 MiB audio buffer directly, avoiding a full multipart `Buffer.concat` copy.
- Gemini prompt streaming yields after the first 66 bytes of a 179-byte SSE response, before response end; legacy buffered response was 75 bytes before first yield.
- Extension lifecycle harness prunes 1000 fired timeouts and 1000 fired RAFs to 0 tracked one-shot handles and clears scoped listeners/timers on unmount/re-evaluation.
- whisper.cpp file transcription avoids the legacy eager 70 MiB Node buffer read (`readCalls=0`, `nodeBufferBytes=0`), while buffer-backed providers still read once.

## Stack validation
- Built from clean `origin/main` on `codex/consolidate-lifecycle-media-wave2`.
- Cherry-picked/adapted only the wave-2 lifecycle/media commits, with clean-base harness drift limited to `scripts/lib/ts-import.mjs` stub support and script-runner assertions unrelated to this wave.
- No coordinator notes, prompt files, local playbooks, pnpm lockfile, or generated package-manager metadata are included.

## Replaces
- #560
- #561
- #563
- #564
- #565
- #570

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   claude-review:
+    # Fork PRs cannot receive the OIDC token required by this action.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     # Optional: Filter by PR author
     # if: |
     #   github.event.pull_request.user.login == 'external-contributor' ||
@@ -41,4 +43,3 @@ jobs:
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,40 @@
+name: Test
+
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'scripts/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/test.yml'
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/**'
+      - 'scripts/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.github/workflows/test.yml'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run node test suite
+        run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
 
       - name: Run node test suite
         run: npm test
--- a/scripts/lib/script-command-runner-harness.mjs
+++ b/scripts/lib/script-command-runner-harness.mjs
@@ -1,0 +1,217 @@
+import { build } from 'esbuild';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
+const runnerPath = path.join(root, 'src/main/script-command-runner.ts');
+
+function makeElectronMockSource(appPaths) {
+  return `
+    const paths = ${JSON.stringify(appPaths)};
+    export const app = {
+      getPath(name) {
+        return paths[name] || paths.userData;
+      },
+    };
+  `;
+}
+
+function makeSettingsMockSource(scriptCommandFolders) {
+  return `
+    export function loadSettings() {
+      return { scriptCommandFolders: ${JSON.stringify(scriptCommandFolders)} };
+    }
+  `;
+}
+
+function makeChildProcessMockSource(childProcessKey) {
+  return `
+    const mock = globalThis[${JSON.stringify(childProcessKey)}];
+
+    export function spawn(...args) {
+      return mock.spawn(...args);
+    }
+  `;
+}
+
+function makeInstrumentedFsSource(metricsKey) {
+  return `
+    import realFs from 'node:fs';
+
+    const metrics = globalThis[${JSON.stringify(metricsKey)}];
+
+    function countReadFile(value, options) {
+      if (Buffer.isBuffer(value)) return value.byteLength;
+      const encoding = typeof options === 'string'
+        ? options
+        : options && typeof options === 'object' && options.encoding
+          ? options.encoding
+          : 'utf8';
+      return Buffer.byteLength(String(value), encoding);
+    }
+
+    export const constants = realFs.constants;
+    export const existsSync = realFs.existsSync.bind(realFs);
+    export const mkdirSync = realFs.mkdirSync.bind(realFs);
+    export const readdirSync = realFs.readdirSync.bind(realFs);
+    export const writeFileSync = realFs.writeFileSync.bind(realFs);
+    export const chmodSync = realFs.chmodSync.bind(realFs);
+    export const unlinkSync = realFs.unlinkSync.bind(realFs);
+    export const openSync = (...args) => {
+      metrics.openSyncCalls += 1;
+      return realFs.openSync(...args);
+    };
+    export const closeSync = realFs.closeSync.bind(realFs);
+    export const accessSync = realFs.accessSync.bind(realFs);
+    export const readFileSync = (filePath, options) => {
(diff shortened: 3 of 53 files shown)


shobhit99:
Reviewed the full diff. Five of the six consolidated changes are clean, behavior-preserving wins; the sixth (the largest file, `ExtensionView.tsx`) ships ~341 lines of production-unreachable code, so its headline benefit is not actually delivered.

## Strengths
- **`script-command-runner.ts`** — moving `clearTimeout` into `finalize()` (L533, L543-546) is a genuine fix: the output-limit overflow path calls `finalize()` directly, so the timeout handle is now cleared there too instead of relying solely on the `close`/`error` listeners. No regression (clearing an already-fired timer is a no-op).
- **`ai-provider.ts` Gemini** — switching `streamGemini` to `:streamGenerateContent?alt=sse` and routing both prompt + chat through the shared `streamGeminiSSE` (with `throwOnNoText`) preserves the no-text/finishReason error semantics while streaming. `readResponseBody` is fully removed with no remaining references. Real time-to-first-byte win.
- **`ai-provider.ts` `transcribeAudio`** — writing multipart parts individually via `writeBufferedRequestParts` and reusing the audio `Buffer` directly instead of `Buffer.concat`-ing an 8 MiB+ body is byte-identical, keeps the same `Content-Length`, and avoids a large copy. Nice.
- **`main.ts` menubar cache** — dedupes title/tooltip/menu/icon native calls; the native menu is only rebuilt when the serialized `items` JSON changes; file-backed icons always refresh (`hasFileBackedMenuBarNativeIcon`), preserving correctness; the "never clear a non-empty tooltip" semantics are preserved. Title logic correctly reads the cached `updateState.lastResolvedTrayIconOk` on icon-skipped ticks. Solid.
- **`whisper-transcribe-file.ts`** — clean DI extraction; the lazy `shouldReadWhisperFileAudioBuffer` read genuinely avoids buffering the audio file for whisper.cpp/native, and `native` early-returns `''` before the API-key checks, matching the original resolution-chain behavior.

## Concerns / potential bugs
- **`ExtensionView.tsx` — the new lifecycle scope is dead code in production (blocking).** The ExtensionView diff is a *single hunk* (`@@ -3362,29 +3362,370 @@`) that only redefines `TimerRegistry`/`createTimerRegistry`/`clearTimerRegistry` and adds `createExtensionLifecycleScope` + `createScopedHostProxy` + `trackEventListener`/`untrackEventListener` + the one-shot-pruning `trackTimeout`/`trackRaf`. **None of these new functions are called by production code.** `loadExtensionExport` and its wrapper invocation (`ExtensionView.tsx` ~L3900-4008 on base) are unchanged: the extension still receives the real `globalThis` (no `scopedWindow`/`scopedDocument`), and still uses the *old* inline `trackTimeout`/`trackRaf` (~L3927-3944) that neither prune fired one-shots nor populate `timeoutClearers`/`rafClearers`. A repo-wide grep finds zero references to `createExtensionLifecycleScope`/`scopedWindow`/`trackEventListener` outside the definitions and `test-extension-lifecycle-sandbox.mjs`.
  - Consequence 1: extensions still call the real `window.addEventListener` directly, so `registry.eventListeners` is always empty and the new `removeEventListener` loop in `clearTimerRegistry` is a no-op — the advertised "clears scoped listeners on unmount" does not happen at runtime.
  - Consequence 2: fired one-shot timeouts/RAFs still accumulate in `registry.timeouts`/`registry.rafs` for the extension's lifetime (the exact "churn" the PR claims to fix) because the production `trackTimeout` isn't the pruning one.
  - The passing `test-extension-lifecycle-sandbox.mjs` exercises `createExtensionLifecycleScope` in isolation, which gives false confidence that the fix is wired. Net: no functional regression (fallback clearers keep existing unmount cleanup working), but the PR's headline extension lifecycle/media benefit is not delivered, and 341 lines of unreachable code + misleading test coverage would ship.
  - Fix: either wire `createExtensionLifecycleScope` into `loadExtensionExport` (pass `scopedWindow`/`scopedDocument` into the `fn(...)` call and swap the inline timer trackers for the scope's), or drop the ExtensionView portion from this consolidation until the wiring lands.

## Suggestions
- After the menubar destructure change, `iconEmoji` (and possibly other locals) may now be read only via `data.*` inside the cache helpers — worth a quick check for now-unused destructured locals, though `tsc -p tsconfig.main.json` reportedly passes.
- This overlaps the #530/#558/#622 lifecycle+media efforts; the non-ExtensionView pieces here are independently valuable and could land even if the extension-scope work is deferred.

🔴 Needs changes before merge — the five main-process changes are safe and worth landing, but the `ExtensionView.tsx` lifecycle scope must be wired into `loadExtensionExport` or removed; as-is it is inert and its tests imply a fix that does not take effect.

JustYannicc:
Addressed the maintainer feedback on the inert ExtensionView lifecycle scope.

What changed:
- Wired `createExtensionLifecycleScope` into production `loadExtensionExport`.
- The bundle wrapper now receives `scopedWindow` for `global`/`globalThis`/`window`/`self`, `scopedDocument` for `document`, and the lifecycle scope's timer APIs for bare `setTimeout`/`setInterval`/`requestAnimationFrame` calls.
- Removed the old inline production timer wrappers that did not prune fired one-shots.
- Hardened timer/RAF callback `this` binding and string timeout evaluation so they stay in the scoped window/document context.
- Updated `scripts/test-extension-lifecycle-sandbox.mjs` so it asserts production wrapper wiring in addition to isolated lifecycle-scope behavior.

Review-thread state checked via GraphQL: PR #575 has 0 reviewThreads, 0 reviews, and 1 top-level maintainer comment; this addresses that actionable comment. The menubar `iconEmoji` suggestion was checked too: `iconEmoji` is still used in `src/main/main.ts`.

Verification:
- `node --test scripts/test-ai-provider-gemini-prompt-streaming.mjs scripts/test-ai-provider-whisper-upload.mjs scripts/test-extension-lifecycle-sandbox.mjs scripts/test-menubar-native-update-cache.mjs scripts/test-script-command-runner.mjs scripts/test-whisper-file-buffer-read.mjs` ✅ 35 passed
- `git diff --check` ✅

JustYannicc:
Follow-up system attribution and amendment (2026-07-10): a full-quit A/B on the affected Mac showed WindowServer staying around 68–73% CPU and MenuBarAgent around 18–22% with SuperCmd completely absent. After relaunch and startup settlement, SuperCmd itself was 0.0–0.1% while the compositor/menu-bar load persisted. This does not support SuperCmd as the cause of the system-wide volume/menu-bar stutter. However, the standalone PR still had one avoidable SuperCmd-side path: file-backed tray icons were decoded/resized and sent through setImage on every accepted title tick. Commit e529dcb adds file identity tracking (path + mtime + size), skips unchanged native image refreshes, and still refreshes when the icon file or image inputs change. Validation: focused menubar test 8/8, full PR test command 64/64, main TypeScript check passed, and git diff check passed.

PR #583, opened 2026-07-04
fix: scope Raycast frecency storage

JustYannicc:
## What changed
- Added extension-scoped frecency storage keys: `sc-frecency:<extensionName>:<namespace>`.
- Reused the existing scoped JSON migration path so legacy `sc-frecency-<namespace>` rankings are copied into the scoped key on first read.
- Extended the focused frecency hook test harness to load the real storage-scope/context runtime and verify scoped-key behavior.

## Why
- The old `sc-frecency-<namespace>` key could collide across Raycast extensions that reused the same namespace.
- Keeping legacy migration preserves existing learned rankings while moving future writes to extension-scoped storage.

## Compatibility impact
- Public `useFrecencySorting` API behavior is unchanged.
- Existing legacy rankings are migrated without deletion or ranking loss.
- No pruning is implemented: dropping retained entries would change semantics for items that disappear and later reappear, because those items are expected to keep their previous ranking data.

## How tested
- `node --test scripts/test-use-frecency-sorting.mjs`
- `npm run typecheck:renderer`
- Codex LSP diagnostics on `src/renderer/src/raycast-api/hooks/use-frecency-sorting.ts` and `src/renderer/src/raycast-api/hooks/storage-scope.ts`.

## Performance evidence
- Focused test fixture starts with the legacy key retaining 2 entries / 102 serialized bytes.
- Legacy migration copies the full payload to each extension scoped key without mutating the legacy payload.
- After an `alpha-extension` visit, the alpha scoped key retains 3 entries / 150 serialized bytes while `bravo-extension` remains isolated.
- After a `bravo-extension` visit to an existing migrated entry, the bravo scoped key retains 2 entries / 102 serialized bytes.

## Stack validation
- Started from required base commit `206eab8280a2eb91a57e4e74ea6df99e45a55cb4` on branch `codex/perf-wave3-integration-stack-20260704` equivalence, then created `codex/perf-wave4-frecency-scoped-storage`.
- Verified the base commit is an ancestor of this branch before implementation.

## Replaces
- Nothing; this is the wave 4 frecency scoped-storage child implementation.


JustYannicc:
Superseded by the clean consolidated wave4 renderer/Raycast PR: #599. That branch contains this PR's covered change on current main without the inherited broad-stack noise.

PR #609, opened 2026-07-05
perf(runtime): cancel extension HTTP work on abort

JustYannicc:
## What changed
- Exposed `cancelHttpRequest(requestId)` through preload and the renderer Electron API type.
- Preserved existing `httpRequest` and one-argument `httpDownloadBinary(url)` compatibility while allowing an optional request ID for cancelable binary downloads.
- Kept the extension fetch abort listener active through binary follow-up downloads, then removed it on every terminal path.
- Reused the main-process `http-request-cancel` map for canceling Node HTTP(S) binary downloads and curl fallback processes.
- Extended the extension fetch abort harness with a binary follow-up abort case.

## Why
The renderer bridge already generated request IDs and attempted to call `electronBridge.cancelHttpRequest`, while main already listened for `http-request-cancel`. The preload/type surface did not expose that method, and binary follow-up downloads did not receive a request ID, so a fetch for binary content could outlive abort/unmount and resolve stale data after the text probe was canceled.

## Compatibility impact
Raycast-compatible call shapes are preserved: existing callers can still call `httpRequest(options)` and `httpDownloadBinary(url)`. The optional request ID is only used by SuperCmd extension fetch cancellation, and non-http/FormData fallback behavior is unchanged.

## How tested
- `node scripts/test-extension-fetch-proxy-abort.mjs`
- `node scripts/test-use-fetch-stale-runs.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json --pretty false`
- LSP diagnostics: no errors in `src/renderer/src/extension-fetch-bridge.ts`, `src/main/preload.ts`, `src/renderer/types/electron.d.ts`, or `src/main/main.ts`.
- `npm run typecheck:renderer` and `npm run build:main` were blocked by the local machine npm policy/parity guard, so direct `tsc` binaries from a temporary sibling `node_modules` symlink were used instead. The symlink was removed before commit.

## Performance evidence
- Before: running the updated binary abort harness against exact base SHA `ac90ce403aeddab98d0a1b245afce268c48b8bfd` failed because the old binary follow-up received `undefined` instead of the active `extensionFetch:*` request ID.
- After: `node scripts/test-extension-fetch-proxy-abort.mjs` passes 4/4, including binary follow-up cancellation, single cancel emission, listener cleanup, and stale byte rejection.

## Stack validation
- Started from `codex/perf-wave4-integrated-stack-20260705` at exact SHA `ac90ce403aeddab98d0a1b245afce268c48b8bfd`.
- Working branch: `codex/perf-runtime-cancel-extension-http-20260705`.
- `git diff --check` passed.

## Replaces
None.


JustYannicc:
Superseded by the consolidated runtime/lifecycle/media PR: https://github.com/SuperCmdLabs/SuperCmd/pull/622

PR #613, opened 2026-07-05
perf(hooks): stabilize Raycast hook signatures

JustYannicc:
## What changed
- Added a shared Raycast hook args signature helper that catches circular data and caches repeat signatures for same-identity args.
- Switched `usePromise` and `useCachedPromise` to the shared stable args helper, removing the direct render-time `JSON.stringify(args || [])` crash path in `useCachedPromise`.
- Keyed `useExec` revalidation on command, argv, and execution-affecting options so dynamic hook inputs trigger execution again.
- Added focused hook regression coverage and a small signature measurement script.

## Why
`useCachedPromise` could stringify hook args directly during render, which made large arg arrays hot and crashed on circular args. `useExec` also passed an empty dependency args list into `usePromise`, so command, args, or execution option changes could be missed until manual revalidation.

## Compatibility impact
Raycast hook API shapes and exports are unchanged. Callback and option refs continue flowing through the existing `usePromise` ref path. `useSQL` already keys dynamic database/query inputs, and `useFetch` already has a guarded request key. `useStreamJSON` option-signature invalidation was left as a non-actionable candidate for this PR to avoid broadening behavior beyond the requested hook invalidation fix.

## How tested
- `node --test scripts/test-raycast-hook-signatures.mjs`
- `node --test scripts/test-raycast-hook-signatures.mjs scripts/test-abortable-promise-hooks.mjs scripts/test-use-fetch-stale-runs.mjs scripts/test-use-stream-json-lifecycle.mjs`
- `node scripts/run-node-tests.mjs --exclude-perf` passed with 317 passing, 1 skipped, 0 failures.
- Codex LSP diagnostics reported no errors for `use-stable-args.ts`, `use-promise.ts`, `use-cached-promise.ts`, and `use-exec.ts`.

## Performance evidence
`node scripts/measure-raycast-hook-signatures.mjs` with 500 iterations over 10,000 large rows:
- direct repeated `JSON.stringify`: 463.667 ms
- cached `getStableArgsKey` repeated same identity: 0.061 ms
- cyclic args signature generation: 2.13 ms without throwing
- measured same-identity speedup: 7601.1x

## Stack validation
Started from `codex/perf-wave4-integrated-stack-20260705` base SHA `ac90ce403aeddab98d0a1b245afce268c48b8bfd` and created `codex/perf-hooks-raycast-runtime-signatures` from that exact commit. The local `npm` command is intentionally disabled on this Mac, so dependencies were populated with runtime pnpm for validation only; `node scripts/check-package-manager-parity.mjs` cannot pass in this environment because that install resolved several package versions differently than `package-lock.json` expects. No dependency files are included in the PR.

## Replaces
N/A


JustYannicc:
Superseded by consolidated renderer/Raycast wave5 PR #621: https://github.com/SuperCmdLabs/SuperCmd/pull/621

PR #619, opened 2026-07-05
perf(renderer): reduce Raycast list render churn

JustYannicc:
## What changed
- Memoized `ListItemRenderer` and `ListEmojiGridItemRenderer` with render-relevant comparators that ignore inline selection/action callback identity.
- Added current-selection/current-primary-action refs in `list-runtime.tsx` so memoized rows keep activation behavior current even when DOM handlers are not refreshed on unrelated selection moves.
- Added `scripts/measure-raycast-list-render-churn.mjs` and `scripts/test-list-emoji-render-churn.mjs` to measure and assert list/emoji selection render churn.

## Why
Selection movement in `list-runtime.tsx` creates fresh row/cell callbacks for every visible list item and emoji cell. Without memoized leaf renderers, every visible row/cell re-rendered even though only the previously selected and newly selected items changed visually.

## Compatibility impact
No Raycast API surface changes. The comparators preserve visible/render-relevant props including selected state, data index, list text values, `assetsPath`, icon references, and accessory references. Callback identity is intentionally ignored, with activation behavior reading the latest selection/action through refs.

## How tested
- `node --test scripts/test-list-emoji-render-churn.mjs scripts/test-list-conditional-virtual-rows.mjs scripts/test-emoji-grid-virtualization.mjs`
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false`
- Codex LSP diagnostics on `src/renderer/src/raycast-api/list-runtime-renderers.tsx`: no diagnostics found
- Codex LSP diagnostics on `src/renderer/src/raycast-api/list-runtime.tsx`: no diagnostics found

## Performance evidence
Before this change, with `SUPERCMD_NODE_MODULES=/Users/justyannicc/.codex/worktrees/3847/SuperCmd/node_modules node scripts/measure-raycast-list-render-churn.mjs --iterations=5 --warmups=1 --selection-steps=800 --json`:
- `selectionRenderChurn.listItemRenderCount`: `51200`
- `selectionRenderChurn.emojiGridItemRenderCount`: `89600`
- `listMedianDurationMs`: `566.0315`
- `emojiMedianDurationMs`: `417.1664`

After this change, with `node scripts/measure-raycast-list-render-churn.mjs --iterations=5 --warmups=1 --selection-steps=800 --json`:
- `selectionRenderChurn.listItemRenderCount`: `1600`
- `selectionRenderChurn.emojiGridItemRenderCount`: `1600`
- `listMedianDurationMs`: `86.1857`
[...]

PR #618, opened 2026-07-05
perf(list): reduce row render churn

JustYannicc:
## What changed
- Memoized `ListItemRenderer` and `ListEmojiGridItemRenderer` with focused comparators that track visible/data props and ignore handler identity churn.
- Added `scripts/measure-raycast-list-render-churn.mjs` to instrument list and emoji-row render churn.
- Added `scripts/test-list-renderer-memoization.mjs` covering comparator behavior and select/activate/context-menu callback wiring.

## Why
Selection, action overlay, scroll, and other parent state changes can recreate visible row handlers even when row visuals do not change. Before this change, that made every visible list row or emoji cell re-enter render. The grid path already had this memoization pattern, so list rows now match that lower-churn behavior.

## Compatibility impact
No Raycast API behavior or extension-facing props were removed or changed. Selection styling, data indexes, assets path icon rendering, accessories, activation, hover selection, and context-menu behavior remain wired through the same props.

## How tested
- `node --test scripts/test-list-renderer-memoization.mjs`
- `node --test scripts/test-list-renderer-memoization.mjs scripts/test-list-registry-stable-order.mjs`
- `node scripts/run-node-tests.mjs --exclude-perf` (317 passed, 1 skipped)
- `./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false`
- `./node_modules/.bin/vite build`
- LSP diagnostics on `src/renderer/src/raycast-api/list-runtime-renderers.tsx`: no diagnostics found

## Performance evidence
Benchmark command: `node scripts/measure-raycast-list-render-churn.mjs --visible-rows=80 --visible-emoji-cells=96 --selection-steps=300 --parent-updates=150 --iterations=5 --warmups=1`

Before:
- list selection churn: 24000 renders, median 293.661ms
- list parent churn: 12000 renders, median 138.918ms
- emoji selection churn: 28800 renders, median 126.492ms
- emoji parent churn: 14400 renders, median 63.981ms

After:
- list selection churn: 600 renders, median 43.205ms
- list parent churn: 0 renders, median 15.695ms
- emoji selection churn: 600 renders, median 34.449ms
- emoji parent churn: 0 renders, median 15.965ms

## Stack validation
[...]

PR #616, opened 2026-07-05
perf(menubar): reuse serialized items for title ticks

JustYannicc:
## What changed
- Reuse the renderer's serialized MenuBarExtra item payload and action map when the registry version and assets path are unchanged.
- Keep tray title, tooltip, and tray icon updates flowing while avoiding stable menu item walks and icon serialization.
- Cache the visible-payload item hash when the serialized item array identity is stable.
- Add focused regression coverage for title-only ticks, handler-only registry updates, and item-hash reuse.

## Why
Title-only menu-bar ticks were still sorting and serializing the full registered item tree, resolving every item and alternate icon, and hashing the full item payload before discovering that only tray-visible fields changed.

## Compatibility impact
Registry, handler, submenu, alternate, section, disabled-label, icon, and extension-context behavior is preserved. Registry version or assets-path changes invalidate the item/action cache, and actions are still installed before unchanged visible payloads are skipped.

## How tested
- `node --test scripts/test-menubar-payload-cache.mjs scripts/test-menubar-renderer-serialization-cache.mjs`
- `node scripts/run-node-tests.mjs --exclude-perf`
- `PATH=/opt/homebrew/lib/node_modules/npm/bin:$PATH node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run typecheck:renderer`
- `PATH=/opt/homebrew/lib/node_modules/npm/bin:$PATH node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:renderer`
- Codex LSP diagnostics on `menubar-runtime-parent.tsx` and `menubar-runtime-payload-cache.ts`: no errors

## Performance evidence
Actual-module baseline before edits, 60 stable items with 12 alternates over 120 title ticks: 7,200 item serializer entries, 8,640 item icon payload calls, 120 tray icon payload calls, 120 IPC sends, 148.81 ms in the timer-based harness.

After edits with the same timer-based harness: 0 item serializer entries, 0 item icon payload calls, 120 tray icon payload calls, 120 IPC sends, 155.35 ms. That wall-clock sample is timer-harness dominated, so I also ran a direct `syncMenuBar`-awaited comparison in the same process: cache-disabled shape 14.61 ms vs current cache 4.25 ms for the same 120 ticks, with item icon calls dropping from 8,640 to 0.

## Stack validation
[...]

PR #611, opened 2026-07-05
perf(raycast): reduce action and dropdown churn

JustYannicc:
## What changed
- Memoized List.Dropdown child flattening while preserving nested Section/Item walking and initial onChange behavior.
- Added action registry visible-field signatures for id/order/title/section/icon/shortcut/style so visible changes publish while execute-only callback changes update in place without new action arrays.
- Kept List, Form, and Detail global keydown listeners attached once and routed them through refs for current actions and shortcut helpers.
- Added focused node tests with before/after instrumentation for dropdown walks, registry publishes, and listener add/remove churn.

## Why
Repeated equivalent renders were rebuilding dropdown item arrays, publishing action arrays for non-visible updates, and reattaching global key listeners when action arrays changed.

## Compatibility impact
No Raycast API behavior or extension compatibility changes intended. Nested dropdown sections/items, initial dropdown onChange, Meta-K, custom shortcuts, primary action execution, and focus restoration semantics are preserved. Grid was validated as non-actionable for global keydown listener churn because it does not attach a global keydown listener.

## How tested
- node --test scripts/test-action-registry-listener-churn.mjs scripts/test-list-dropdown-memoization.mjs
- node --test scripts/test-registry-dirty-checks.mjs scripts/test-list-registry-stable-order.mjs scripts/test-registry-microtask-lifecycle.mjs
- npm run test:ci
- node_modules/.bin/vite build
- LSP diagnostics on changed TSX runtime files: no diagnostics found
- LSP references for new exported helpers resolved only to intended local/tested uses

Note: npm run typecheck:renderer is blocked on this machine by /Users/justyannicc/.local/bin/npm, which prints "npm is intentionally disabled on this Mac. Using pnpm instead." and trips the repo package-manager parity guard. Direct tsc -p tsconfig.renderer.json --noEmit --pretty false runs but reports pre-existing project-wide renderer Node global type errors for process/Buffer outside this change.

## Performance evidence
- List.Dropdown stable children over 8 renders: child walks 40 before -> 5 after; initial onChange calls 8 before -> 1 after.
- Action registry repeated equivalent execute-only renders: registry version publishes 12 before -> 0 after; visible style/shortcut/icon updates still publish.
- Global keydown equivalent action churn over 12 renders: List/Form/Detail addEventListener 12 -> 1 and removeEventListener 11 -> 0; Grid remains 0/0 because it has no global keydown listener.

## Stack validation
[...]

PR #621, opened 2026-07-05
perf(renderer): consolidate Raycast UI hot paths

JustYannicc:
## What changed
- Consolidated renderer/Raycast/menubar wave5 performance work into one branch from current upstream `origin/main`.
- Reduced selected-scroll churn for Raycast Grid and launcher list selected-row correction by using instant scroll behavior.
- Reduced Raycast action churn with visible-field action signatures and stable global keydown listeners; kept List.Dropdown flattening cached for stable children without relying on it as a broad rerender win.
- Stabilized Raycast hook argument signatures for `usePromise`, `useCachedPromise`, and `useExec`.
- Reduced List and emoji-grid render churn with memoized renderers and virtual-row helpers.
- Reduced MenuBarExtra title tick churn by reusing serialized renderer menu items and skipping unchanged native tray image updates for stable file-backed icons.

## Why
The replaced micro PRs target the same renderer/Raycast UI hot paths and were hard to review independently because several changes overlap in list rendering, action registry updates, hook signatures, and menubar title tick behavior. This PR keeps those changes together while avoiding unrelated runtime, command, search, and validation stack changes.

## Compatibility impact
Raycast API shapes and exports are preserved. Selection semantics, action execution, dropdown initial `onChange`, shortcut handling, menu item actions, tray title fallback behavior, file-backed icon refresh on mtime/size changes, and native menu serialization remain compatible.

## How tested
- PASS: dependency hydration with real Homebrew npm from committed `package-lock.json`: `/opt/homebrew/Cellar/node@22/22.22.3/bin/node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js ci --ignore-scripts --force`; no lockfile churn committed.
- PASS: standard `npm test` path with real npm: `/opt/homebrew/Cellar/node@22/22.22.3/bin/node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js test` (`node --test 'scripts/test-*.mjs'`), 38 tests total, 37 passed, 1 live Electron test skipped.
- Covered by standard `npm test` path (`node --test 'scripts/test-*.mjs'`): `scripts/test-virtual-selection-scroll-behavior.mjs`, `scripts/test-action-registry-listener-churn.mjs`, `scripts/test-list-dropdown-memoization.mjs`, `scripts/test-raycast-hook-signatures.mjs`, `scripts/test-list-renderer-memoization.mjs`, `scripts/test-list-emoji-render-churn.mjs`, `scripts/test-menubar-native-update-cache.mjs`, and `scripts/test-menubar-renderer-serialization-cache.mjs`.
- Manual performance evidence harnesses, intentionally outside the `test-*.mjs` glob: `scripts/measure-raycast-hook-signatures.mjs` and `scripts/measure-raycast-list-render-churn.mjs`.
- PASS: `/opt/homebrew/Cellar/node@22/22.22.3/bin/node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:main`.
- PASS: `/opt/homebrew/Cellar/node@22/22.22.3/bin/node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:renderer`.
- PASS: `git diff --check`.
- PASS: `node --check` over all added/touched `.mjs` scripts.
- PASS: Codex LSP diagnostics on every touched TS/TSX file found no errors; remaining diagnostics were hints only.

## Performance evidence
- Manual list structural measurement: `node scripts/measure-raycast-list-render-churn.mjs --iterations=2 --warmups=1 --selection-steps=200 --json` produced `selectionSteps=200`, `listItemRenderCount=400`, `emojiGridItemRenderCount=400`, median about `0.0014ms`, p95 about `0.0014ms`.
- Manual hook signature measurement: `node scripts/measure-raycast-hook-signatures.mjs --iterations=2 --warmups=1 --args=500 --json` produced cached stable-args speedup about `14020x` versus repeated direct `JSON.stringify`, and handled cyclic args.
- Source micro PR evidence retained: #611 reported action registry execute-only publishes `12 -> 0` and global keydown listener adds `12 -> 1`; its dropdown harness still covers stable-children flattening `40 -> 5`, but that case is intentionally scoped.
- Source micro PR evidence retained: #613 reported cached stable hook args avoiding repeated large `JSON.stringify` work and handling cyclic args without throwing.
- Source micro PR evidence retained: #610 reported stable file-backed title ticks reducing native tray icon resolves `60 -> 1` and `setImage` calls `59 -> 0`.

## Stack validation
- Branch `codex/consolidate-wave5-renderer-raycast` was created/reset from `SuperCmdLabs/SuperCmd:main` at `9bdd3d2`.
- Source micro PR branches were not merged wholesale because they were based on a broader wave4 integration stack.
- Intentional same-domain prerequisite subset: list virtual-row helpers from #533 were ported because #619 depends on them; unrelated #533 icon/native/package/launcher changes were not included, so #533 is not superseded by this PR.
- Final diff scope is renderer/Raycast/menubar plus focused scripts: `src/renderer/src/raycast-api/*`, `src/renderer/src/App.tsx`, `src/main/main.ts`, and focused `scripts/*` harnesses.

## Replaces
- #604
- #610
- #611
- #613
- #616
- #618
- #619

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   claude-review:
+    # Fork PRs cannot receive the OIDC token required by this action.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     # Optional: Filter by PR author
     # if: |
     #   github.event.pull_request.user.login == 'external-contributor' ||
@@ -41,4 +43,3 @@ jobs:
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
 
       - name: Run node test suite
         run: npm test
--- a/scripts/measure-raycast-hook-signatures.mjs
+++ b/scripts/measure-raycast-hook-signatures.mjs
@@ -1,0 +1,94 @@
+#!/usr/bin/env node
+
+import fs from 'node:fs';
+import path from 'node:path';
+import vm from 'node:vm';
+import { createRequire } from 'node:module';
+import { performance } from 'node:perf_hooks';
+
+const require = createRequire(import.meta.url);
+const ts = require('typescript');
+const helperPath = path.resolve('src/renderer/src/raycast-api/hooks/use-stable-args.ts');
+
+function loadHelper() {
+  const source = fs.readFileSync(helperPath, 'utf8');
+  const transpiled = ts.transpileModule(source, {
+    compilerOptions: {
+      module: ts.ModuleKind.CommonJS,
+      target: ts.ScriptTarget.ES2022,
+      esModuleInterop: true,
+      importsNotUsedAsValues: ts.ImportsNotUsedAsValues.Remove,
+    },
+    fileName: helperPath,
+  });
+
+  const module = { exports: {} };
+  const sandbox = {
+    Array,
+    Error,
+    JSON,
+    Map,
+    Math,
+    module,
+    Object,
+    require: (request) => {
+      if (request === 'react') {
+        return { useRef: (value) => ({ current: value }) };
+      }
+      return require(request);
+    },
+    String,
+    Symbol,
+    WeakMap,
+    WeakSet,
+    exports: module.exports,
+  };
+
+  vm.runInNewContext(transpiled.outputText, sandbox, { filename: helperPath });
+  return module.exports;
+}
+
+function measure(label, fn) {
+  const start = performance.now();
+  const result = fn();
+  const durationMs = performance.now() - start;
+  return { label, durationMs: Number(durationMs.toFixed(3)), result };
+}
+
+const { getStableArgsKey } = loadHelper();
+const rows = Array.from({ length: 10000 }, (_, index) => ({
+  index,
+  label: `row-${index}`,
+  payload: 'x'.repeat(64),
+}));
+const largeArgs = [rows];
+const cyclic = { rows };
+cyclic.self = cyclic;
+
+const iterations = Number(process.env.SUPERCMD_HOOK_SIGNATURE_ITERATIONS || 500);
+const direct = measure('direct JSON.stringify repeated', () => {
+  let key = '';
+  for (let index = 0; index < iterations; index += 1) {
+    key = JSON.stringify(largeArgs);
+  }
+  return key.length;
+});
+
+getStableArgsKey(largeArgs);
+const cached = measure('getStableArgsKey repeated same identity', () => {
+  let key = '';
+  for (let index = 0; index < iterations; index += 1) {
+    key = getStableArgsKey(largeArgs);
+  }
+  return key.length;
+});
+
+const cyclicResult = measure('getStableArgsKey cyclic args', () => getStableArgsKey([cyclic]).length);
+
+console.log(JSON.stringify({
+  iterations,
+  direct,
+  cached,
+  cyclicResult,
+  cachedSpeedup: Number((direct.durationMs / Math.max(cached.durationMs, 0.001)).toFixed(2)),
+}, null, 2));
--- a/scripts/measure-raycast-list-render-churn.mjs
+++ b/scripts/measure-raycast-list-render-churn.mjs
@@ -1,20 +1,18 @@
 #!/usr/bin/env node
 
-import fs from 'node:fs/promises';
+import fs from 'node:fs';
 import path from 'node:path';
 import { performance } from 'node:perf_hooks';
-import { pathToFileURL, fileURLToPath } from 'node:url';
-import * as esbuild from 'esbuild';
+import { fileURLToPath } from 'node:url';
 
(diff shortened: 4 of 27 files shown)


shobhit99:
Reviewed the full diff (main.ts tray caching, App.tsx, action registry, detail/form/list/grid runtimes, the hook stable-args extraction, the list virtual-row helpers + emoji-grid virtualization, and the menubar serialization cache). Focused on behavior/API preservation for the Raycast UI runtime. Net: solid, mostly a correctness-neutral-or-better perf pass; one low-severity behavioral nuance in emoji-grid scrolling.

**Strengths**
- **Memoized row renderers are safe.** `ListItemRenderer`/`ListEmojiGridItemRenderer` are wrapped in `React.memo` with explicit comparators that track visible props (title/subtitle/icon/accessories/isSelected/dataIdx/assetsPath) and deliberately ignore handler identity. The reason this doesn't break click/activate is that `onActivate` now reads `selectedIdxRef.current`/`primaryActionRef.current` (both refreshed every render), so a memoized-stale closure still executes against the current selection/action. `key={item.id}` means the comparator only ever runs for the same item, so `id` doesn't need comparing. Correct.
- **Action registry publish gate is a superset of the old one.** `buildActionVisibleSignature` includes icon/shortcut/style/order in addition to title/sectionTitle, so it publishes in strictly more cases than the previous `id:title:sectionTitle` snapshot (the old code actually *missed* icon/shortcut/style changes). `execute` is always mutated in place, so the newest callback fires even when no re-publish happens. `unregister` always publishes. Verified against `ActionRegistration` in `action-runtime-types.ts` — all Pick'd fields exist.
- **Install-once global keydown effects** (list/form/detail) with `[]` deps reading from a per-render-updated ref eliminate add/removeEventListener churn with no stale closures. The list ref omits `primaryAction` correctly (its global handler only does metaK + shortcut iteration; Enter is handled in the React `onKeyDown`).
- **useExec fix:** passing real `execArgs` instead of `[]` fixes a latent staleness bug where `useExec` never re-ran on command/args/options change. Shared `useStableArgs`/`getStableArgsKey` is circular-ref-safe and behavior-equivalent to the removed private copy (confirmed no other importer of the old one).
- **main.ts tray caching is sound.** Refresh is gated on a composite icon key + a `statSync` mtime/size file-identity key, which correctly covers the "icon file became available with the same path" case (`|missing` → `|mtimeMs=…|bytes=…` flips the key). `lastResolvedTrayIconOk` is restored/persisted across skipped refreshes so the `⏱` title-fallback state is preserved; state is cleaned up on `menubar-remove`.
- **Menubar serialization cache is consistent.** Keyed on `(registryVersion, assetsPath)`; since every register/unregister bumps `registryVersion`, the cache can never diverge from `registryRef`, and action closures read `runtimeCtxRef.current` at call time, so no stale-callback risk. Good test coverage added throughout.

**Concerns / potential bugs**
- **(Low) Emoji-grid scroll-into-view can silently no-op on large selection jumps** — `list-runtime.tsx`, the emoji branch of the selection scroll effect (`if (shouldUseEmojiGridValue) { el.querySelector('[data-idx="…"]')?.scrollIntoView(...) }`). Now that the emoji grid is virtualized, only rows within the overscan window are in the DOM. Incremental arrow nav is fine (target stays within overscan). But a jump beyond the window — e.g. `selectedIdx` clamped to `filteredItems.length - 1` when a filter shrinks the list, or a controlled selection far off-screen — finds no element and doesn't scroll, leaving the selection off-viewport until manual scroll. Not strictly a regression (base applied 36px linear-row math to the grid layout, which was also wrong), but the linear-list path already computes the offset from `rowMetrics`/`itemIdxToRowIdx` and calls `el.scrollTo` regardless of render window; the emoji path could use the same math for consistency instead of `querySelector`.

**Suggestions**
- App.tsx has an unrelated `command?.name` → `command?.title` swap in the `[PIN-TOGGLE]` debug log plus two `as React.Dispatch<…>` casts — harmless but off-theme for a perf PR; if `CommandInfo` keys on `name`, that log now prints `undefined`. Worth confirming intent or dropping.
- `flattenListDropdownItems` is memoized on `[children]`; React children are usually fresh each render, so the memo will rarely hit in practice (behavior is still preserved — the initial `onChange` fires once via the `didEmitInitialChange` ref). Just a smaller win than the description implies.
- Large diff (~2k lines, mostly new test/measure scripts) folding in several micro-PRs (#604/#610/#611/#613/#616/#618/#619, plus a subset of #533's virtual-row helpers). Source changes are well-isolated per file, but note some bisect risk from the consolidation.

🟡 Approve with minor comments

JustYannicc:
Addressed the review feedback in commit `8ec499c`.

What changed:
- Fixed the virtualized emoji-grid selection scroll path in `list-runtime.tsx`: selection scroll now uses the existing virtual row map/row offsets and `scrollTo(...)`, so large jumps no longer depend on the selected emoji cell already being mounted in the DOM.
- Extended `test-virtual-selection-scroll-behavior.mjs` to guard that the Raycast list selection path uses offset-based instant scroll and does not fall back to `scrollIntoView`/mounted-cell lookup.
- Dropped the incidental App.tsx `[PIN-TOGGLE]` debug logs and removed the off-theme setter casts; fixed `useAiChat`'s return type so the real state setter type flows without casts.
- Softened the PR body's List.Dropdown memoization wording to scope the evidence to stable children instead of presenting it as a broad rerender win.

Verification:
- `node --test scripts/test-action-registry-listener-churn.mjs scripts/test-list-dropdown-memoization.mjs scripts/test-list-renderer-memoization.mjs scripts/test-raycast-hook-signatures.mjs scripts/test-virtual-selection-scroll-behavior.mjs` ✅ 8 passed
- `git diff --check` ✅

PR #614, opened 2026-07-05
perf(media): cancel binary fetch downloads

JustYannicc:
## What changed
- Extended `http-download-binary` with an optional `{ requestId }` payload while preserving `httpDownloadBinary(url)` callers.
- Reused the existing HTTP cancel IPC path for binary downloads, including Node transport cancel, curl fallback cancel, redirect handoff, active request map cleanup, and bounded curl stderr in composed errors.
- Wired extension fetch binary GETs and the axios binary shim to pass request IDs and cancel on abort/timeout.
- Added focused lifecycle harnesses for binary cancellation, active request map cleanup, URL-only compatibility, and axios timeout cleanup.

## Why
Binary GETs could outlive renderer aborts/timeouts and keep buffering in main after the renderer had already stopped waiting. The text fetch proxy already had request IDs and cancellation; the binary transport needed the same behavior-preserving lifecycle path.

## Compatibility impact
- Existing `window.electron.httpDownloadBinary(url)` and Raycast-style binary download callers continue to work.
- Binary return shape remains a full `Uint8Array`; no public streaming API was introduced.
- Redirects, headers, curl fallback, axios `arraybuffer`/`stream` behavior, and extension-observable response semantics are preserved.

## How tested
- `node scripts/test-extension-fetch-proxy-abort.mjs`
- `node scripts/test-http-download-binary-cancel.mjs`
- `node scripts/test-camera-capture-single-encode.mjs`
- `node scripts/test-cloud-media-buffering.mjs`
- `node scripts/test-http-response-decode-async.mjs`
- `node scripts/test-use-fetch-stale-runs.mjs && node scripts/test-raycast-ai-abort-listeners.mjs`
- `node scripts/check-package-manager-parity.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit --pretty false && ./node_modules/.bin/tsc -p tsconfig.renderer.json --noEmit --pretty false`
- Codex LSP diagnostics: no errors for `src/main/main.ts`, `src/main/preload.ts`, `src/renderer/src/extension-fetch-bridge.ts`, `src/renderer/src/ExtensionView.tsx`, and `src/renderer/types/electron.d.ts`.

## Performance evidence
- Before: renderer-side timeout/abort could leave `http-download-binary` running in main, retaining chunk buffers until completion/error.
- After: canceling by request ID destroys the active Node request or curl child, clears retained chunks, deletes the active request map entry in `finally`, and clears axios timeout/abort listeners.
[...]

PR #612, opened 2026-07-05
perf(ipc): harden native helper readiness

JustYannicc:
## What changed
- Replaced 50 ms polling startup readiness waits for Whisper.cpp, Parakeet, Qwen3, and audio capturer with event-driven readiness waiters backed by one startup timeout.
- Added shared native-helper output buffer utilities and applied conservative caps to JSON-line stdout buffers plus plain stderr/stdout buffers found during the audit.
- Extended the native helper lifecycle harness to cover timer count, no-poll readiness, bounded buffer behavior, pending request rejection cleanup, and stale-child regressions.

## Why
Malformed native helper output could grow partial-line buffers without bound, and readiness polling allocated repeating timers while waiting for events the helpers already emit.

## Compatibility impact
- Preserves existing native helper JSON line protocol semantics for normal ready/result/error messages.
- Preserves timeout, killed-process, superseded-child, and pending-request rejection behavior.
- Leaves generic spawn IPC and legacy Raycast-compatible channels unchanged.

## How tested
- `node --test scripts/test-native-helper-process-lifecycle.mjs` passed: 8 tests.
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit --pretty false` passed.
- Codex LSP diagnostics on `src/main/main.ts` and `src/main/native-helper-lifecycle.ts`: no diagnostics found.
- `git diff --check` passed.

## Performance evidence
- Startup readiness wait test asserts one `setTimeout` and zero `setInterval` allocations, replacing the previous 50 ms polling interval per helper startup.
- Ready-event settlement clears the single startup timeout immediately when the helper emits ready.
- Native helper partial-line buffers are capped at 256 KiB, and oversized completed malformed lines are dropped while normal JSON lines continue through.

## Stack validation
- Started from verified base SHA `ac90ce403aeddab98d0a1b245afce268c48b8bfd` on `codex/perf-wave4-integrated-stack-20260705`.
- Built implementation branch `codex/perf-ipc-native-helper-lifecycle`.
- Confirmed no tracked package metadata changes from local dependency installation.

## Replaces
- Supersedes the wave5 delegated implementation request for native helper readiness, buffer bounds, and subprocess lifecycle cleanup.
- Rejected/deferred generic spawn IPC event batching because the existing path already emits unified `spawn-event` plus legacy channels, and batching would broaden compatibility/timing risk beyond this scoped helper fix.

PR #607, opened 2026-07-05
perf(background): prevent overlapping refresh ticks

JustYannicc:
## What changed
- Added a shared per-timer in-flight guard for background refresh ticks.
- Wrapped extension and inline-script interval callbacks so a timer skips ticks while its previous async refresh is still running.
- Added regression coverage for overlapping extension refresh, overlapping inline script refresh, and independent timers running concurrently.

## Why
Slow `runExtension`, hidden no-view enqueueing, `runScriptCommand`, or `fetchCommands` work could outlive the configured interval. The previous `setInterval(async () => ...)` path allowed repeated overlapping starts for the same background command, creating avoidable IPC/script/build pressure over long sessions.

## Compatibility impact
- Manual launches are unaffected because only background interval callbacks use the new guard.
- Independent background commands still run independently because each timer gets its own guard closure.
- Menu-bar active checks remain inside the extension background tick before dispatching `sc-launch-extension-bundle`, preserving the existing Raycast-compatible menu-bar behavior.

## How tested
- `node --test scripts/test-background-refresh-inflight.mjs`
- `node --test scripts/test-background-refresh-timer-diff.mjs scripts/test-background-no-view-dedupe.mjs`
- `node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run typecheck:renderer`
- `node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js run build:renderer`
- Codex LSP diagnostics on `src/renderer/src/hooks/backgroundRefreshTimers.ts` and `src/renderer/src/hooks/useBackgroundRefresh.ts`: no errors.

## Performance evidence
Focused slow-work timer harness, `intervalMs=10`, `workMs=35`, `durationMs=130`:
- Before edits: `totalStarts=10`, `completed=10`, `maxConcurrent=4`.
- After edits through the production guard: `totalStarts=3`, `completed=3`, `maxConcurrent=1`.

## Stack validation
Started from `codex/perf-wave4-integrated-stack-20260705` at `ac90ce403aeddab98d0a1b245afce268c48b8bfd`, verified with `git rev-parse HEAD` before editing. Implementation branch: `codex/perf-background-refresh-inflight-20260705`.

## Replaces
None.

PR #606, opened 2026-07-05
perf(extensions): reuse on-demand build stamps

JustYannicc:
## What changed
- Shared extension build input/options creation between full extension builds and on-demand single-command builds.
- On successful on-demand builds, write a `.sc-build-stamp.json` command entry with the same input/output signature shape used by `buildAllCommands`.
- Merge existing command stamps only when the extension-level build context still matches, then replace the on-demand command stamp with the fresh signature.
- Extended the multi-command build harness with an on-demand single build followed by a full build reuse case.

## Why
- Wave 4 stamps let full builds reuse unchanged bundles, but on-demand `buildSingleCommand` did not write a stamp.
- A command built on demand could be rebuilt by the next full build even when source and config were unchanged.

## Compatibility impact
- Preserves Raycast-compatible externals, native scheme handling, tsconfig handling, dependency installation fallback, missing-source behavior, and full rebuild recovery paths.
- No manifest or runtime API changes.

## How tested
- `node scripts/test-extension-runner-multi-command-build.mjs`
- `SUPERCMD_EXTENSION_BUILD_MEASURE=1 node scripts/test-extension-runner-multi-command-build.mjs`
- `node scripts/test-extension-bundle-cache.mjs`
- `node scripts/measure-extension-bundle-cache.mjs`
- `./node_modules/.bin/tsc -p tsconfig.main.json --noEmit --pretty false`
- Codex LSP diagnostics for `src/main/extension-runner.ts`: no diagnostics found.

## Performance evidence
- New harness scenario `on-demand-single-then-full-build`: on-demand build esbuild entries `[`command-0`]; follow-up full build esbuild entries `[`command-1`, `command-2`]`, proving the on-demand-stamped command was reused instead of rebuilt.
- Representative measured run: single `15.06ms`, follow-up full `13.53ms`; esbuild call durations `13.57ms` for `command-0` and `12.63ms` for `command-1`/`command-2`.
- Existing repeated full-build measurement remains one esbuild call per fixture after warm reuse: command counts 1/5/20 reported timings `[13.82, 0.33, 0.2]`, `[17.85, 0.58, 0.51]`, `[18.86, 1.71, 1.81]` with `esbuildCalls: 1`.

## Stack validation
- Started from `codex/perf-wave4-integrated-stack-20260705` exact SHA `ac90ce403aeddab98d0a1b245afce268c48b8bfd`; branch now descends from that SHA.
- Branch: `codex/perf-extensions-on-demand-build-stamps-20260705`.
[...]

PR #622, opened 2026-07-05
perf(runtime): consolidate lifecycle and media hot paths

JustYannicc:
## What changed
- Consolidates the selected runtime/lifecycle/media/native-helper performance PRs into one domain-scoped branch from `origin/main`.
- Adds bounded AI stream parser buffers, async HTTP response decode, cloud media streaming, extension command refresh/build reuse, scoped extension lifecycle timers/listeners, guarded background refresh ticks, abortable extension HTTP/binary downloads, and native-helper readiness/buffer lifecycle hardening.
- Adds standard regression tests under `scripts/test-*.mjs` for the consolidated behavior.

## Why
- The wave5 micro PRs depend on prior same-domain runtime consolidation that is not present on upstream `main`.
- Consolidating the runtime/lifecycle/media/native-helper lane avoids a fragile stack of dependent PRs while keeping unrelated renderer/commands/search work out of scope.

## Compatibility impact
- Preserves Raycast runtime compatibility and existing extension API behavior.
- Keeps URL-only binary download callers working while adding request-id cancellation support.
- Keeps extension command build outputs compatible while reusing valid build stamps and cached refresh state.

## How tested
- PASS: direct standard regression command `node --test 'scripts/test-*.mjs'`: 84 passed, 2 skipped.
- Covered by standard `npm test` because `package.json` defines it as `node --test 'scripts/test-*.mjs'`: `scripts/test-ai-provider-stream-parser-bounds.mjs`, `scripts/test-background-refresh-inflight.mjs`, `scripts/test-cloud-media-buffering.mjs`, `scripts/test-extension-command-refresh-cache.mjs`, `scripts/test-extension-fetch-proxy-abort.mjs`, `scripts/test-extension-lifecycle-sandbox.mjs`, `scripts/test-extension-runner-multi-command-build.mjs`, `scripts/test-http-download-binary-cancel.mjs`, `scripts/test-http-response-decode-async.mjs`, `scripts/test-native-helper-process-lifecycle.mjs`, `scripts/test-registry-microtask-lifecycle.mjs`.
- PASS: `node node_modules/typescript/bin/tsc -p tsconfig.main.json`.
- PASS: `node node_modules/vite/bin/vite.js build`.
- PASS: `git diff --check origin/main...HEAD`.
- PASS: codex-lsp diagnostics on touched TS/TSX files.
- Local npm wrapper note: the wrapper did not reach the test script because npm is disabled on this machine and the pnpm install shim stopped on ignored build scripts; the direct command used by `npm test` passed.
- Non-gating native-build note: `node scripts/build-native.mjs` could not resolve `node-addon-api` from this workspace install. No native source was touched by this consolidation.

## Performance evidence
- Standard regression/performance guard tests include parser buffer bounds, async decompression timer delay, media streaming, command refresh/build reuse, lifecycle cleanup, refresh inflight suppression, and cancellation cleanup.
- Manual measurement harness, intentionally outside `npm test`: `scripts/measure-extension-bundle-cache.mjs`; local run reported repeated bundle-cache scenarios with 20 iterations in roughly 3.063ms and 2.73ms.

## Stack validation
- Branch: `codex/consolidate-runtime-lifecycle-final`, from current `origin/main`, pushed to `JustYannicc`.
- Target: `SuperCmdLabs/SuperCmd:main`.
- Diff scope inspected with `git diff --stat origin/main...HEAD` and `git diff --name-status origin/main...HEAD`; touched files stay in runtime/lifecycle/media/native-helper scope plus associated Raycast runtime hooks and proof tests.
- Did not include broad unrelated renderer, commands/search, or integration-stack changes beyond the selected same-domain prerequisite consolidation and wave5 PRs.

## Replaces
- #601 `perf(runtime): consolidate extension lifecycle and media fixes`
- #603 `perf(ai): bound stream parser buffers`
- #605 `perf(lifecycle): scope extension timer builtins`
- #606 `perf(extensions): reuse on-demand build stamps`
- #607 `perf(background): prevent overlapping refresh ticks`
- #609 `perf(runtime): cancel extension HTTP work on abort`
- #612 `perf(ipc): harden native helper readiness`
- #614 `perf(media): cancel binary fetch downloads`

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   claude-review:
+    # Fork PRs cannot receive the OIDC token required by this action.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     # Optional: Filter by PR author
     # if: |
     #   github.event.pull_request.user.login == 'external-contributor' ||
@@ -41,4 +43,3 @@ jobs:
           prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }}'
           # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
           # or https://code.claude.com/docs/en/cli-reference for available options
-
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           cache: npm
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --force
 
       - name: Run node test suite
         run: npm test
--- a/scripts/lib/ts-import.mjs
+++ b/scripts/lib/ts-import.mjs
@@ -2,15 +2,44 @@
 // esbuild on the fly. This lets the recovery tests run the *real* production
 // source (renderer-recovery.ts, reload-budget.ts) instead of grepping it.
 //
-// Only works for modules with no relative imports of their own — the recovery
-// helpers are deliberately dependency-free for exactly this reason.
+// By default this stays intentionally lightweight for dependency-free helpers.
+// Tests that need production modules with a few heavy dependencies can pass
+// stubs keyed by import specifier.
 
 import { transform } from 'esbuild';
 import fs from 'node:fs';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
 
-export async function importTs(absPath) {
+function toDataUrl(code) {
+  return 'data:text/javascript;base64,' + Buffer.from(code).toString('base64');
+}
+
+async function transformTs(source) {
+  const { code } = await transform(source, { loader: 'ts', format: 'esm' });
+  return code;
+}
+
+function rewriteImportSpecifiers(code, absPath, options) {
+  const stubs = options.stubs || {};
+  const stubUrls = new Map(Object.keys(stubs).map((specifier) => [specifier, toDataUrl(stubs[specifier])]));
+
+  const rewriteSpecifier = (specifier) => {
+    if (stubUrls.has(specifier)) return stubUrls.get(specifier);
+    if ((specifier.startsWith('./') || specifier.startsWith('../')) && options.root) {
+      const resolved = path.resolve(path.dirname(absPath), specifier);
+      return pathToFileURL(resolved).href;
+    }
+    return specifier;
+  };
+
+  return code
+    .replace(/(\bfrom\s*["'])([^"']+)(["'])/g, (_match, prefix, specifier, suffix) => `${prefix}${rewriteSpecifier(specifier)}${suffix}`)
+    .replace(/(\bimport\s*["'])([^"']+)(["'])/g, (_match, prefix, specifier, suffix) => `${prefix}${rewriteSpecifier(specifier)}${suffix}`);
+}
+
+export async function importTs(absPath, options = {}) {
   const src = fs.readFileSync(absPath, 'utf8');
-  const { code } = await transform(src, { loader: 'ts', format: 'esm' });
-  const dataUrl = 'data:text/javascript;base64,' + Buffer.from(code).toString('base64');
-  return import(dataUrl);
+  const code = rewriteImportSpecifiers(await transformTs(src), absPath, options);
+  return import(toDataUrl(code));
 }
--- a/scripts/measure-extension-bundle-cache.mjs
+++ b/scripts/measure-extension-bundle-cache.mjs
@@ -1,0 +1,258 @@
+#!/usr/bin/env node
+
+import { fileURLToPath, pathToFileURL } from 'node:url';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { createRequire } from 'node:module';
+import { build } from 'esbuild';
+import { performance } from 'node:perf_hooks';
+
+const require = createRequire(import.meta.url);
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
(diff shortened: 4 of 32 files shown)


shobhit99:
### Review: perf(runtime): consolidate lifecycle and media hot paths

Large consolidation (6.4k+ / 785-) folding several wave PRs (per the body, replaces #601/#603/#605/#606/#607/#609/#612/#614). Reviewed on its own merits. Most of it is solid leak-hardening and event-driven refactors, but I found one change that looks like an unintended and serious behavior inversion for extension `require()` of Node builtins.

**Strengths**
- Native-helper readiness (`native-helper-lifecycle.ts`) replaces the 50ms `setInterval` polling with an event-driven `markReady`/timeout wait, and every `kill*Server`/`exit` handler now guards `process === child` to avoid a superseded restart clobbering a newer process. Genuinely better.
- HTTP request/download cancellation + abort-listener cleanup (`ai-provider.ts` `httpRequest`/`transcribeAudio`, `main.ts` http-request/http-download-binary, ollama-pull `finishPullRequest`) consistently de-dupe settlement and remove abort listeners on end/close/error — real leak fixes.
- `spawn-process` per-sender tracking with cleanup on `render-process-gone`/`destroyed` is a good backstop against orphaned child processes on renderer crash.
- Registry microtask guards (`action-/grid-/list-/menubar-runtime`) add a `mountedRef` so queued `queueMicrotask` updates and post-unmount `register/unregister` calls no-op — prevents stale setState.
- `useBackgroundRefresh` key-based reconcile + `createInFlightBackgroundRefreshTick` (skip overlapping ticks) is a clean improvement; cleanup-on-unmount effect is correct.
- Streaming ElevenLabs TTS to disk via `pipeline` (instead of buffering the whole audio) with unlink-on-failure is a good media hot-path win.

**Concerns / potential bugs**

1. 🔴 **Builtin `require` priority inverted — extensions now get weak stubs instead of real Node.** `ExtensionView.tsx`: the new `shouldUseSuperCmdBuiltinFacade()` gate runs *before* `tryRealNodeRequire()` and returns `nodeBuiltinStubs[name]` for **any** name in that map (fs, crypto, zlib, http, https, os, path, stream, util, net, tls, dns, …). Base behavior explicitly "Prefer real Node … better than returning a stub" (`USE_REAL_NODE_BUILTINS = true`), so this flips resolution from real → stub for all of them. That matters because several stubs are non-functional: `cryptoStub.createHash(...).digest('hex')` returns `Math.random().toString(16)` (a **random string, not a hash** — `ExtensionView.tsx:1329-1333`), `createCipheriv/Decipheriv` return empty buffers, and `zlib.gzipSync/gunzipSync` are identity no-ops (`ExtensionView.tsx:2680-2688`). Concrete failure: an extension doing `crypto.createHash('sha256').update(data).digest('hex')` (integrity/cache keys, HMAC API signing) silently gets garbage; `zlib` decompression returns raw bytes. The lifecycle goal only requires facading `child_process` (for spawn tracking) and `timers` (for scoping) — the gate should be narrowed to those, letting everything else fall through to `tryRealNodeRequire` as before. The added tests don't cover builtin fidelity, so this passes CI. Please confirm whether routing all builtins through stubs was intended; if not, this is a broad extension-compat regression.

2. 🟡 **Native-helper line buffer silently drops valid long lines.** `native-helper-lifecycle.ts` `appendNativeHelperLineBuffer`: a *complete* line with `line.length >= maxChars` (256KB) is discarded (`continue`) rather than emitted. The genuine flood protection is the incomplete-buffer `slice(-maxChars)`; dropping a completed newline-terminated line means a large-but-valid JSON response from a helper (e.g. a long qwen3/transcription result serialized on one line) is thrown away and the pending request hangs until timeout instead of resolving. Consider only bounding the *incomplete* tail and letting oversized complete lines through (or failing the pending request explicitly rather than dropping).

3. 🟡 **Scoped `window`/`document` Proxy is a broad semantic change riding in a "perf" PR.** `createExtensionLifecycleScope` now hands extensions Proxy-wrapped `window`/`document` (previously raw `globalThis`). The isolation/cleanup intent is good and `getOwnPropertyDescriptor` forcing `configurable:true` is the standard invariant workaround, but this changes identity/descriptor semantics for every extension and is hard to bisect within a 6k-line diff. Worth calling out as the main regression-surface risk; heavier real-extension smoke testing (not just the sandbox unit tests) would de-risk it before merge.

**Suggestions**
- Given the size, consider splitting the `window`/`document` scoping change out from the pure leak/cancellation fixes so a bad interaction can be reverted independently.

Verdict: 🔴 Needs changes before merge — concern #1 (crypto/zlib/etc. stub inversion) should be resolved or explicitly justified; #2/#3 are worth addressing.

JustYannicc:
Addressed the review feedback in `f85bee3`.

What changed:
- Builtin `require()` priority: narrowed the SuperCmd builtin facade gate to `child_process` only. Timer modules still use the scoped lifecycle facade path, while `crypto`, `zlib`, `fs`, etc. now fall through to real Node before compatibility stubs. Added `scripts/test-extension-builtin-require-priority.mjs` to lock this in.
- Native-helper long lines: `appendNativeHelperLineBuffer` no longer drops newline-terminated complete lines based on length; only the incomplete trailing buffer remains capped. Updated native-helper lifecycle coverage with an oversized complete JSON line case.
- Window/document proxy scoping: kept the lifecycle scoping change, added commentary in `createScopedHostProxy`, and expanded sandbox coverage for scoped identity, `window`/`self`/`globalThis`, `document.defaultView`, descriptor invariants, enumeration, write-through, host method binding, and cleanup.

Verification:
- `node --test scripts/test-extension-builtin-require-priority.mjs scripts/test-native-helper-process-lifecycle.mjs scripts/test-extension-lifecycle-sandbox.mjs` ✅ 21 passed
- `npm run build:main` ✅
- `git diff --check` ✅

PR #467, opened 2026-06-10
feat(whisper): custom vocabulary biasing for local whisper.cpp

Arturgrr:
## What

Adds a **Custom Vocabulary** field to **Settings → AI → Whisper** that feeds whisper.cpp's native `initial_prompt` parameter. This biases recognition toward user-provided terms (e.g. `Kotlin`, `Electron`, `Raycast`) that the local `ggml-base` model frequently mis-transcribes.

Same mechanism Superwhisper uses for its vocabulary feature. Scoped to the local `whispercpp` provider — cloud providers are out of scope for this PR.

## Why

Technical terms and proper nouns are often spelled wrong by the base model. Until now the `initial_prompt` field of `whisper_full_params` was left empty, even though the bundled whisper.cpp already supports it. This wires up the existing capability and exposes it in the UI.

## Changes

- **Swift** (`whisper-transcriber.swift`): thread `initialPrompt` into `transcribeWithContext` via nested `withCString` (keeps the C pointer valid for the whole `whisper_full` call); parse `initial_prompt` in the persistent server loop.
- **Main** (`main.ts`): read `s.ai.speechVocabulary` in the `whisper-transcribe` IPC handler and forward it through `transcribeAudioWithWhisperCpp` (no renderer/preload changes needed — vocabulary comes from settings).
- **Settings** (`settings-store.ts`, `electron.d.ts`): new `speechVocabulary` string field (default `""`).
- **UI** (`AITab.tsx`, `index.css`): vocabulary textarea in the whispercpp block with a soft ~224-token warning + counter (no hard limit; whisper.cpp truncates internally). Adds a `--status-warning` theme token (light + dark).
- **i18n**: `settings.ai.whisper.vocabulary.*` keys added to `en.json` and translated across all 9 locales (parity verified: 0 missing / 0 extra).

## Notes

- `initial_prompt` is a probabilistic bias, not a hard dictionary — it favors the listed spellings but doesn't guarantee them. Capped at ~224 tokens by whisper.cpp.
- Backward compatible: empty vocabulary → prompt not set → identical behavior to before.

## Verification

- `swiftc` compile of the transcriber: ✅ exit 0
- `tsc` main process: ✅ clean
- `tsc` renderer: ✅ no new errors introduced
- i18n structural parity across 9 locales: ✅ `missing: 0, extra: 0`
[...]

PR #305, opened 2026-04-21
Bug fixes apr 21

shobhit99:

PR #145, opened 2026-03-18
Complete app language support across settings and launcher surfaces

eachann1024:
## Summary

This PR completes the first end-to-end app language pass for SuperCmd across both the Settings window and the launcher chrome.

It keeps the original settings localization work, carries it through the latest upstream release merge, follows maintainer feedback by moving the app language selector into Advanced settings and switching it to a dropdown, adds Russian locale support, and now makes System Settings panes prefer macOS-localized bundle titles instead of relying on a manually maintained translation list.

Related issues

None.

## Motivation

The branch originally added app language selection and localized settings surfaces, but several launcher strings were still hard-coded in English and the language control was placed in General settings. After the latest upstream release landed, the branch also needed a clean merge resolution so the localization work could still be reviewed and tested on top of the current main branch.

This update closes those gaps by keeping the branch current with upstream, finishing the remaining launcher-facing translations, aligning the language selector UX with maintainer feedback, and letting System Settings titles follow the localized names exposed by macOS itself. That avoids growing a second hand-maintained translation table just for Apple-owned preference panes.

## Features Added

- Localize the launcher search placeholder, grouped section headers, footer labels, and command action labels.
- Show localized built-in command titles such as the settings entry instead of always rendering the English system title.
- Move the app language selector from General to Advanced settings and switch it from pills to a dropdown.
- Add Russian locale support across runtime locale resolution, persisted settings, language options, and launcher or settings-facing translations.
- Resolve System Settings pane titles from macOS-localized bundle metadata and resource fallbacks before falling back to the raw bundle name.
- Complete `settings.ai` locale coverage for German, French, and Spanish so those catalogs no longer depend on English fallback for that section.

## Changes

- `src/renderer/src/App.tsx` and `src/renderer/src/utils/command-helpers.tsx`:
  - Routed launcher placeholder text, section labels, footer text, calculator prompts, action labels, and built-in command titles through the shared translation runtime.
- `src/renderer/src/settings/AITab.tsx`:
  - Resolved the upstream release merge conflict while keeping the localization layer compatible with the new AI settings changes.
- `src/renderer/src/settings/GeneralTab.tsx` and `src/renderer/src/settings/AdvancedTab.tsx`:
  - Removed the language pills from General and reintroduced the setting in Advanced with a dropdown styled consistently with the Whisper model selector.
- `src/renderer/src/i18n/runtime.ts`, `src/main/settings-store.ts`, and `src/renderer/types/electron.d.ts`:
[...]

PR #636, opened 2026-08-03
feat(i18n): add Polish (pl) as an app display language

jash90:
## What changed

Adds Polish (`pl`) as a selectable app display language, alongside the ten locales already supported. Follows the exact pattern used when Russian (#145) and Italian (`202109b`, #305) were added.

**New file**
- `src/renderer/src/i18n/locales/pl.json` — full Polish translation of `en.json` (982 keys).

**Wiring**
- `src/renderer/src/i18n/runtime.ts` — `pl` added to the import list, `SupportedAppLocale`, `APP_LANGUAGE_OPTIONS`, `MESSAGE_CATALOG`, `normalizeAppLanguage()` and `resolveAppLocale()`.
- `src/main/settings-store.ts` — `pl` added to the `AppLanguage` union **and** to the main-process copy of `normalizeAppLanguage()`. This second one is easy to miss: without it the renderer sends `appLanguage: 'pl'`, the main process normalises it back to `'system'` on both the load path and the patch path, and the picker silently snaps back after a restart.
- `src/renderer/types/electron.d.ts` — `appLanguage` union widened to match.

**Display name**
- `settings.general.language.pl` added to all 10 existing locale files, in each target language (`Polnisch`, `Polonais`, `Polaco`, `Polacco`, `Польский`, `ポーランド語`, `폴란드어`, `波兰语`, `波蘭語`). That matches the convention of the two most recent additions (`ru`, `it`), which are localized exonyms; the older entries in `fr`/`de`/`es`/`ko` are a mix of endonyms and exonyms and were left untouched.

**Docs**
- `docs/i18n-architecture.md` and `CLAUDE.md` still listed the supported locales as ending at `es` / `ru` — both were already missing `it`, and the doc was also missing `ru`. Refreshed to the actual current set, including the inline parity-check snippet in `CLAUDE.md`.

No TSX changes were needed: `AdvancedTab.tsx` already renders `APP_LANGUAGE_OPTIONS.map(...)` with a dynamic `t('settings.general.language.' + option)` lookup.

## Why

The app already ships Polish **dictation** (`pl-PL`, added in #442), so the UI language was the missing half. Adding a locale here is meant to be a resource-only change — which is exactly the stated goal in `docs/i18n-architecture.md`.

## Compatibility impact

- **No behaviour change for existing users.** No existing key, string, or default was modified. `DEFAULT_APP_LANGUAGE` is still `'system'`.
- **Settings stay forward/backward compatible.** An older build reading a `settings.json` containing `"appLanguage": "pl"` falls through `normalizeAppLanguage()` to `'system'` — the existing, intended behaviour for unknown values. No migration needed.
- **Users on a `pl-PL` system locale with `appLanguage: 'system'` will now see Polish instead of English.** That is the intended behaviour of every locale addition in this repo, and it is opt-out via the picker.
- No new dependency and no bundle-config change — locale JSON is an ordinary ESM import bundled by Vite into `dist/renderer`, already covered by the `"dist/**/*"` glob in `build.files`.

## How I tested

```
npm run check:i18n                                 # [i18n] pl: OK
SUPERCMD_STRICT_LOCALES=ko,pl npm run check:i18n   # exit=0
tsc -p tsconfig.main.json --noEmit                 # clean
tsc -p tsconfig.renderer.json --noEmit             # no new errors (see note)
npm run build:renderer                             # built in 6.30s
npm test                                           # 81 pass, 0 fail, 1 skipped
```

Also verified the Polish strings actually land in the bundle (`grep "Jak w systemie" dist/renderer/assets/*.js`).

Two honest notes:

1. `tsc -p tsconfig.renderer.json` reports **44 errors on a clean checkout of `main`**, before this change. I diffed the error sets with and without my commit — they are identical apart from one pre-existing error in `runtime.ts` (`Type alias 'MessageTree' circularly references itself`) shifting from line 15 to 16 because of the added import. This PR introduces **zero** new type errors. There is no `typecheck` script and CI does not run `tsc`, so this is pre-existing.
2. `npm run check:i18n` also prints a **pre-existing** warning for `it` (missing `settings.ai.whisper.vocabulary.*` and `settings.extensions.appSearchScope.*`, left behind by #466 and #467). Untouched here — I deliberately left it out to keep this PR to one concern, and I'm happy to send it as a separate PR if useful.

I could not run `npm run build:native` / `npm run dev` end-to-end in this environment: `npm ci` fails on Apple Silicon because `@esbuild/darwin-x64` is a direct (non-optional) dependency and npm rejects it with `EBADPLATFORM` on `arm64`. That is the same issue #635 is fixing; I worked around it with `npm ci --ignore-scripts --force`. So the runtime restart-persistence check on `settings-store.ts` is reasoned-through rather than clicked-through — worth a reviewer's second look if you have a working dev setup.

## Translation quality — for a reviewer who doesn't read Polish

I'm a native Polish speaker and this is hand-translated from `en.json` only — no file was copied from another locale as a base (per `CLAUDE.md` rule 6). Register follows Apple's Polish macOS conventions: nominal section titles, imperative buttons, impersonal `-no/-to` forms for completed actions (`Zapisano`, `Odinstalowano`) so nothing is gendered, infinitive questions for confirmations (`Odinstalować „{name}"?`), and sentence case throughout — Polish does not use Title Case. Terminology is anchored on macOS PL: Clipboard→`Schowek`, Shortcut→`Skrót`, Trash→`Kosz`, Appearance→`Wygląd`, tab→`karta` vs bookmark→`zakładka`.

Brand terms are kept verbatim: SuperCmd, Whisper, Parakeet, Qwen3, ggml, CoreML, ElevenLabs, OpenAI, Anthropic, Gemini, Ollama, LM Studio, Raycast, Excalidraw, Supermemory, and all model names.

**One deliberate deviation worth flagging.** Two strings use `{plural}`, which carries English morphology (`""` or `"s"`) — `canvas.count` and `settings.ai.speak.elevenlabs.voice.available`. Polish has three numeral forms (1 / 2–4 / 5+) that the runtime cannot express, and injecting `{plural}` would emit a literal `s`. I dropped **only** that token in those two strings and rephrased to a label-colon-number form that needs no agreement (`Tablice: {count}`). `{count}` is preserved. Every other placeholder in all 982 keys is preserved verbatim — verified mechanically below.

Three checks a non-Polish speaker can run:

**1. Structural parity (must exit 0)**
```bash
SUPERCMD_STRICT_LOCALES=pl npm run check:i18n
```

**2. Placeholder parity per key (reports 0 mismatches; the two `{plural}` drops above are the only allowed exceptions)**
```bash
node -e 'const fs=require("fs"),d="src/renderer/src/i18n/locales/";const f=(o,p="",r={})=>{for(const k of Object.keys(o)){const v=o[k],q=p?p+"."+k:k;v&&typeof v==="object"?f(v,q,r):r[q]=v}return r};const en=f(JSON.parse(fs.readFileSync(d+"en.json","utf8"))),pl=f(JSON.parse(fs.readFileSync(d+"pl.json","utf8")));const A={"canvas.count":["plural"],"settings.ai.speak.elevenlabs.voice.available":["plural"]};const t=s=>[...String(s).matchAll(/\{([^}]+)\}/g)].map(m=>m[1]).sort();let b=0;for(const k in en){const x=t(en[k]),y=t(pl[k]??""),a=A[k]||[];const m=x.filter(i=>!y.includes(i)&&!a.includes(i)),v=y.filter(i=>!x.includes(i));if(m.length||v.length){b++;console.log(k,m,v)}}console.log(b?b+" mismatches":"placeholders OK")'
```

**3. Brand terms preserved (counts should match `en.json`)**
```bash
for t in SuperCmd Whisper Parakeet Raycast Ollama ElevenLabs; do
  printf "%-12s en=%s pl=%s\n" "$t" \
    "$(grep -o "$t" src/renderer/src/i18n/locales/en.json | wc -l | tr -d ' ')" \
    "$(grep -o "$t" src/renderer/src/i18n/locales/pl.json | wc -l | tr -d ' ')"
done
```

For context, here is the state of every locale measured against `en.json`. "Identical" counts values byte-identical to the English one — that number is never 0 for any language, because brand names, API-key placeholders and URLs are supposed to stay as-is. The second column is the one that matters: values that are unambiguously untranslated English *sentences* (contain a space, end in `.`/`!`/`?`).

| locale | identical to EN | untranslated English sentences |
|---|---|---|
| `ru` | 615 | **173** |
| `de` | 522 | **138** |
| `fr` | 519 | **138** |
| `es` | 517 | **138** |
| `it` | 85 (+14 missing keys) | 7 |
| `ja` | 54 | 1 |
| `zh-Hans` | 39 | 1 |
| `zh-Hant` | 41 | 1 |
| `ko` | 49 | 0 |
| **`pl` (this PR)** | 54 | **0** |

`pl.json`'s 54 identical values are exclusively brand/product names (`SuperCmd Whisper`, `Parakeet TDT v3`, `ElevenLabs Scribe v1`, `LM Studio`, …), API-key placeholders (`sk-...`, `client-123`, `https://api.supermemory.ai`), and single technical tokens (`OK`, `Alias`, `Host`, `Enter`, `Hindi`, `AI`, `LLM`) — plus two pure-placeholder strings (`{progress}% · {transferred} / {total}`, `Bang !{bang}`). None of them is an English sentence. Reproduce with:

```bash
node -e 'const fs=require("fs"),d="src/renderer/src/i18n/locales/";const f=(o,p="",r={})=>{for(const k of Object.keys(o)){const v=o[k],q=p?p+"."+k:k;v&&typeof v==="object"?f(v,q,r):r[q]=v}return r};const en=f(JSON.parse(fs.readFileSync(d+"en.json","utf8")));const S=v=>{const s=String(v).trim();return /\s/.test(s)&&/[.!?]$/.test(s)&&!/^https?:/.test(s)};for(const l of ["ru","de","fr","es","it","ja","ko","zh-Hans","zh-Hant","pl"]){const o=f(JSON.parse(fs.readFileSync(d+l+".json","utf8")));const same=Object.keys(en).filter(k=>k in o&&en[k]===o[k]);console.log(l.padEnd(9),"identical:",String(same.length).padStart(3)," sentences:",same.filter(k=>S(en[k])).length)}'
```

https://claude.ai/code/session_01DM3kF3wjMpoDJSUWApqsJ5


jash90:
Heads-up on the two red checks — neither is caused by this PR:

**`test`** fails in `npm ci`, before any of my code is touched:

```
npm error code EBADPLATFORM
npm error notsup Unsupported platform for @esbuild/darwin-arm64@0.19.12:
  wanted {"os":"darwin","cpu":"arm64"} (current: {"os":"linux","cpu":"x64"})
```

`@esbuild/darwin-arm64` and `@esbuild/darwin-x64` are declared as direct (non-optional) dependencies, so `npm ci` cannot resolve them on `ubuntu-latest`. The same failure is on `main` itself — the last `Test` run on `main` (`29134507376`, 11 Jul) died with the identical error. #635 is the fix for this.

Locally on macOS (with `npm ci --ignore-scripts --force` to get past the same platform check on arm64) the suite is green:

```
# tests 82
# pass 81
# fail 0
# skipped 1
```

**`claude-review`** fails because the workflow uses the `pull_request` event with `secrets.CLAUDE_CODE_OAUTH_TOKEN`, and GitHub does not expose repository secrets to workflows triggered from a fork. Expected for any external contribution.

Happy to rebase once #635 lands if you would rather see a green CI run first.

jash90:
Self-review pass — pushed four copy fixes after checking every interpolated string against its actual call site rather than trusting `en.json` alone. Three of them were real defects in my translation.

**1. `settings.ai.whisper.elevenlabsWarning`** — `{action}` is `t("common.add").toLowerCase()` (`AITab.tsx:1395`), not an English word, so in Polish it renders as `dodaj`. My string had it directly after a full stop, producing a lowercase sentence start. Reworded so the imperative sits mid-sentence:
`"Wybrano ElevenLabs STT. W sekcji Klucze API {action} klucz API ElevenLabs."`

**2. `settings.extensions.folder.failed`** — same deal, `{action}` is `dodaj`/`usuń` (`ExtensionsTab.tsx:880,903`). Polish `Nie udało się…` wants an infinitive, but an imperative is injected, so I quote it as the operation name instead of my earlier parenthetical:
`"Nie udało się wykonać operacji „{action}” na folderze rozszerzeń."`

**3. `onboarding.voice.permissionNotes.blocked`** — `{target}` is `Mikrofon` / `Rozpoznawanie mowy` (`OnboardingExtension.tsx:623-629`). My `Dostęp do funkcji {target}` read unnaturally; quoting sidesteps the declension:
`"Uprawnienie „{target}” jest zablokowane. …"`

**4. `store.sections.moreCommands`** — reworded to `+{count} więcej poleceń` to match the English wording.

Re-verified after the change: `check:i18n` → `pl: OK`; `SUPERCMD_STRICT_LOCALES=ko,pl` → exit 0; placeholder parity 0 mismatches; 982/982 keys; `tsc -p tsconfig.main.json` clean; `build:renderer` OK; `npm test` 81 pass / 0 fail.

---

Two **pre-existing** bugs I ran into while doing this. Both are outside the scope of a localization PR so I have not touched them, but they affect every locale:

- **`store.sections.moreCommands` is double-counted.** `StoreTab.tsx:1100` renders `+{ext.commands.length - visibleCommands.length} {t("store.sections.moreCommands")}` while the string itself is `"+{count} more commands"` and is called with no values. So it currently displays `+3 +{count} more commands`. The fix is to pass `{ count }` to `t()` and drop the JSX-side `+{…}` — after which every locale, including `pl`, renders correctly with no further translation change.
- **`elevenlabsWarning` starts a sentence with a lowercased word in English too**: `"ElevenLabs STT selected. add ElevenLabs API key in API Keys."` (because of `.toLowerCase()` on `common.add`). Also affects `de`/`fr`/`ru`, which left the string in English, and `it`, whose `Impossibile {action}` expects an infinitive but receives an imperative.

Happy to send either as a separate PR if you want them.

Issue #634, opened 2026-07-22
Intel Macs: all native Swift helper binaries are arm64-only (bad CPU type) — breaks Cloud Sync, window fine-tune, snippets, and more


jacob-bd:
## System info

- MacBookPro16,1 (Intel), Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
- macOS 26.5.2 (Build 25F84)
- SuperCmd 1.0.26, installed from `SuperCmd-1.0.26.dmg` (the Intel/x64 build, not `-arm64`)

## Summary

On Intel Macs, **every native Swift helper binary bundled with SuperCmd is compiled arm64-only**, so any feature that shells out to one of them fails immediately and unconditionally with "bad CPU type in executable." This is not a permissions, network, or timing issue — it reproduces 100% of the time, on every launch, regardless of environment.

## Root cause

`scripts/build-native.mjs` compiles each Swift helper like this:

```js
run(`swiftc -O -o ${out} ${src} ${frameworks}`);
```

No `-target` flag is passed. Without one, `swiftc` compiles for whatever architecture is running the build. Since release builds are apparently produced on Apple Silicon, every helper in `dist/native/` (`window-adjust`, `snippet-expander`, `settings-coordinator`, `whisper-transcriber`, `calendar-events`, `microphone-access`, `color-picker`, `speech-recognizer`, `screen-ocr`, etc. — the full list in that script) comes out arm64-only. There's no per-architecture build pass and no `lipo` step to produce universal binaries, so this same `dist/native/` directory appears to get bundled into **both** the `-arm64` and the plain (Intel) release packages.

Confirmed directly on the machine above:

```
$ file /Applications/SuperCmd.app/Contents/MacOS/SuperCmd
.../SuperCmd: Mach-O 64-bit executable x86_64

$ file /Applications/SuperCmd.app/Contents/Resources/app.asar.unpacked/dist/native/*
.../settings-coordinator: Mach-O 64-bit executable arm64
.../window-adjust:        Mach-O 64-bit executable arm64
.../snippet-expander:     Mach-O 64-bit executable arm64
... (every single native/ binary is arm64, ~20 of them)

$ /Applications/SuperCmd.app/Contents/Resources/app.asar.unpacked/dist/native/settings-coordinator "/some/path/settings.json"
zsh: bad CPU type in executable: .../settings-coordinator
```

The main app executable is correctly built as x86_64 for this install — only the Swift-compiled helpers are wrong.

[...]

PR #638, opened 2026-08-08
fix(build): compile native Swift helpers as universal arm64+x86_64 binaries

YuriNachos:
## Problem

The 15 native Swift helper binaries built by `scripts/build-native.mjs` (`get-selected-text`, `color-picker`, `screen-ocr`, `snippet-expander`, …) were compiled as single-arch **arm64** executables, so on Intel Macs every feature that shells out to one of them fails with `bad CPU type in executable` (#634). The HyperKey monitor binary that `src/main/main.ts` compiles at runtime had the same single-arch problem.

## Solution

Swift's driver rejects the `-arch` flag, so the usual "compile once with multiple `-arch` flags" path does not work for Swift sources. Instead, `scripts/build-native.mjs` now:

1. Compiles each helper twice — once targeting `arm64-apple-macos11`, once targeting `x86_64-apple-macos11` (the Swift sources themselves are unchanged; only the deployment target and the second arch slice are added).
2. Merges the two slices with `lipo -create` into the final universal (arm64 + x86_64) Mach-O binary.
3. Cleans up the per-arch intermediate files in a `finally`, so a failed arch build doesn't leave stale slices behind.

`ensureHyperKeyMonitorBinary` in `src/main/main.ts` gets the same slice-and-lipo treatment, with one difference: the host-arch slice is mandatory, while the second slice is best-effort — on a toolchain that cannot cross-compile (e.g. an Intel Mac with only Command Line Tools) it falls back to a working host-arch binary instead of disabling HyperKey.

`macos11` (Big Sur) is the deployment floor — the arm64 minimum and a broad floor for Intel Macs.

## Testing

- [x] `npm test` (`node --test scripts/test-*.mjs`) — 96 passed, 0 failed, including the new `scripts/test-native-universal-arch.mjs`, which asserts that each built helper is universal (`lipo -archs` reports both `arm64` and `x86_64`). The new tests skip on non-macOS hosts, where the native binaries are not built and `lipo` does not exist.
- [x] `node scripts/build-native.mjs` (`build:native`) — completes clean; the produced helpers are universal.
- [x] `npm run build` (full `build:main` + `build:renderer` + `build:native`) — green.

## Notes

- The Node native addon (`native_helpers.node`) is unaffected — it is already handled by the existing arch-aware build path; only the Swift helper binaries needed the universal treatment.
- Binaries that come from separate build pipelines (`whisper-transcriber`, `soulver`, `parakeet`) are out of scope here.


YuriNachos:
I've had this one in mind since #463 sorted esbuild's native binary for both architectures — the Swift helpers were the remaining piece building single-arch, and this PR closes that gap for arm64+x86_64. Is there anything you'd like from my side to move it toward review, such as a verified Intel build or a rebase onto the latest main? Whatever helps, I'm on it. If a smaller change is an easier entry point, #639, #640 and #642 are open as well, and I'd gladly prioritise whichever is most useful.

PR #644, opened 2026-08-14
fix(whisper): preserve transcript when insertion fails

### LINKED ISSUE #643: Whisper can silently lose a completed transcript when text insertion fails
## Problem

When SuperCmd Whisper successfully transcribes speech but cannot insert it into the intended target, the resulting text can be silently lost.

This happens when there is no editable text target, focus has moved, or macOS injection/paste fails. The UI reports that it could not type into the active app, but does not preserve the final transcript anywhere useful.

## Why this is a regression

Earlier Whisper code attempted a clipboard fallback when direct paste failed. The current `whisper-type-text-live` path returns `fallbackClipboard: false` on every result path and does not retain the transcript.

Relevant current code:
- `src/main/main.ts`, `whisper-type-text-live`
- `src/renderer/src/SuperCmdWhisper.tsx`, `autoPasteAndClose`

## Expected behaviour

A completed transcription should never be discarded solely because text injection failed.

At minimum, when insertion fails:

1. Copy the final transcript to the system clipboard.
2. Show an actionable message: `Couldn't insert text — copied to clipboard.`
3. Avoid polluting SuperCmd Clipboard History with the temporary internal paste operation, but allow the deliberate failure fallback to be captured normally.

## Nice follow-up

An optional bounded Whisper transcript history would make this much more robust, with timestamp, copy/re-paste actions, and clear retention/privacy controls. That should be a separate feature; the clipboard fallback is the immediate data-loss fix.

## Acceptance criteria

- [ ] Failure to type/paste a non-empty final transcript leaves that transcript on the system clipboard.
- [ ] The user sees a clear recovery message rather than only a generic error.
- [ ] Normal successful dictation restores the pre-existing clipboard as it does today.
- [ ] The fallback path has a regression test.

## Reproduction

1. Invoke hold-to-talk Whisper while no editable field is focused, or make the target unavailable before release.
2. Speak a unique sentence and release.
3. Observe that the transcription is not inserted and cannot be recovered from Clipboard History or Whisper.

mplacona:
## Summary

Fixes #643.

When Whisper finishes transcribing but cannot inject text into its target app, SuperCmd now copies the final transcript to the system clipboard instead of discarding it.

- Adds a small, tested clipboard-preservation helper.
- Returns the existing `fallbackClipboard` signal accurately.
- Keeps the Whisper overlay open and tells the user that the text was copied, rather than closing and losing the recovery message.
- Successful dictation still uses the existing temporary clipboard/paste flow and restores the user's prior clipboard.

## Verification

- `node --test scripts/test-whisper-clipboard-fallback.mjs`
- `npm test` — 83 passed, 0 failed, 1 existing Electron-dependent test skipped
- `npm run build:main`
- `npm run build:renderer`
- `npm run check:i18n` — exit 0; reports two pre-existing missing Italian keys

## Manual test to run on macOS

1. Invoke hold-to-talk Whisper without an editable target or make the target unavailable before release.
2. Dictate a unique phrase.
3. Confirm Whisper remains visible with `Couldn't insert text — copied to clipboard.`
4. Paste into any app and confirm the phrase is recoverable.


mplacona:
Hey, is anyone maintaining this repo?

